PR #8, opened 2023-04-03
Added apt update

thomhug:
Doesn't work with an 'apt update' in between. 
Added a newline at the end :)


thirdeyenick (approved):
Thanks...lgtm

PR #12, opened 2023-05-15
Add multi-stage creation wait

ctrox:
This extends the creator with waitStages for resources which want to wait on any other dependent resources in the whole creation flow. This also reworks the printing system as the previous approach was a bit flawed. The spinner is now on the line that is currently in progress and is then updated to a tick/cross depending on the outcome of the stage.


create/create.go
```diff
@@ -30,59 +28,134 @@ type Cmd struct {
...
+
+func (m *message) printSuccess() {
+	if m.disabled {
+		return
+	}
+
+	format.PrintSuccessf(m.icon, m.text)
+}
+
+func newCreator(client *api.Client, mg resource.Managed, resourceName string) *creator {
```
thirdeyenick, line 67: is this actually needed? Seems to just pass function args to a struct.

ctrox: It's not really needed right now but if we add new fields to `creator` at some point they could be defaulted here without needing to change all call-sites. Also this makes it possible to do something like `newCreator(...).wait(...)`, which you can't do when creating a new struct.

thirdeyenick: Ok, lets leave it in then. I was curious about the last statement. Not sure if I did it correctly, but it seems to work for me: https://go.dev/play/p/7sjk41nKy8C.

ctrox: Ah right, forgot that you can wrap the struct initialization in braces, looks pretty ugly though 😄

thirdeyenick (approved):

PR #13, opened 2023-05-16
Add application create

ctrox:
This adds the full app creation flow. By default it will wait until the application is fully built and released.

The args allow to configure all the settings that are currently available on the API.

The second commit retries the watch on failure. I noticed this during testing with the new API, whenever our staging env kills the watch (for whatever reason) it would fail silently and the wait would just timeout.

Note to the reviewer: Please install this branch locally and try it out against staging `nctl --api-cluster staging.nineapis.ch create app ...`.

### DIFF
--- a/create/application.go
+++ b/create/application.go
@@ -1,0 +1,277 @@
+package create
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	runtimev1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/utils/pointer"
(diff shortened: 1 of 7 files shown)


create/application.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func (b buildError) Error() string {
+	return fmt.Sprintf("build failed with status %s", b.build.Status.AtProvider.BuildStatus)
+}
+
+func (b buildError) printMessage() {
+	fmt.Printf("\nYour build has failed with status %q. To view the build logs run the command:\n",
+		b.build.Status.AtProvider.BuildStatus)
+	fmt.Printf("  nctl logs build %s\n\n", b.build.Name)
```
ctrox, line 138: This is just a placeholder, the command might end up being different. It would also be kind of cool if it could show the last x (10?) log lines of the build as chances are good they might already contain the problem.

twhiston: That would be really cool, if it had last 10 lines plus that call to action to view all the logs it would be a really nice developer experience

thirdeyenick (approved):

create/application.go
ctrox: I added this to the builds CLI ticket, it should be straightforward to add once we have logging built in.

PR #15, opened 2023-05-22
Add git auth flags to app create

ctrox:
These can also be set using env variables as this might be preferred to setting them directly via the CLI flag.

### DIFF
--- a/create/application_test.go
+++ b/create/application_test.go
@@ -21,41 +21,99 @@ func TestApplication(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cmd := applicationCmd{
-		Wait:        false,
-		Name:        "custom-name",
-		GitURL:      "https://github.com/ninech/doesnotexist.git",
-		GitSubPath:  "/my/app",
-		GitRevision: "superbug",
-		Size:        "mini",
-		Hosts:       []string{"custom.example.org", "custom2.example.org"},
-		Port:        1337,
-		Replicas:    42,
-		Env:         map[string]string{"hello": "world"},
-	}
-
-	app := cmd.newApplication("default")
 	client := fake.NewClientBuilder().WithScheme(scheme).Build()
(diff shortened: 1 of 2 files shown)


create/application_test.go
```diff
@@ -47,10 +51,18 @@ func TestApplication(t *testing.T) {
...
 	assert.Equal(t, cmd.Name, app.Name)
-	assert.Equal(t, cmd.GitURL, app.Spec.ForProvider.Git.URL)
-	assert.Equal(t, cmd.GitSubPath, app.Spec.ForProvider.Git.SubPath)
-	assert.Equal(t, cmd.GitRevision, app.Spec.ForProvider.Git.Revision)
+	assert.Equal(t, cmd.Git.URL, app.Spec.ForProvider.Git.URL)
+	assert.Equal(t, cmd.Git.SubPath, app.Spec.ForProvider.Git.SubPath)
+	assert.Equal(t, cmd.Git.Revision, app.Spec.ForProvider.Git.Revision)
+	assert.Equal(t, cmd.Git.Username, string(authSecret.Data["username"]))
+	assert.Equal(t, cmd.Git.Password, string(authSecret.Data["password"]))
+	assert.Equal(t, cmd.Git.SSHPrivateKey, app.Spec.ForProvider.Git.Auth.SSHPrivateKey)
```
pawelkuc, line 65: this line doesn't test anything now. We should also test secret data here, i.e. similar to the un and pw:
```
assert.Equal(t, cmd.Git.SSHPrivateKey, string(authSecret.Data["privatekey"]))
```
of course we should also set some test data for the `privatekey`

pawelkuc (approved):

create/application_test.go
ctrox: Good catch. Though we need to create a separate app to test that since the user/pass and ssh key are mutually exclusive.

pawelkuc: looks good 👌

PR #17, opened 2023-05-23
Add application logs

ctrox:
This commit adds a new log client package that makes use of the loki
logcli client and automatically initializes the client with the token of
the user. For application logs the following is implemented:

* Setting the lookback duration for the logs
* Setting the output format (default and json)
* Limiting the amount of lines returned
* Live tail (follow) of logs


gajicdev (approved):
looks good!!

did you try it with a real loki already?

gajicdev:
checked out your second commit and looks good, approved 👍

PR #19, opened 2023-05-30
Add build commands

ctrox:
This adds a few commands to interact with depl.io builds.

* Get command to list builds and see details with `-o yaml`
* Flag `--pull-image` to pull the build output from the registry to the local docker host
* `logs build` command to view log output of a build

Note to the reviewer: please clone the branch locally and give the new commands a spin using `nctl --api-cluster staging.nineapis.ch`. Logging infra is still missing but build pulls should already work.


get/build.go
```diff
@@ -0,0 +1,128 @@
...
+		printEmptyMessage(apps.BuildKind, client.Namespace)
+		return nil
+	}
+
+	if cmd.PullImage {
+		if len(cmd.Name) == 0 {
+			return fmt.Errorf("build name has to be specified for pulling an image")
+		}
+
+		if err := pullImage(ctx, client, &buildList.Items[0]); err != nil {
```
thirdeyenick, line 54:
```suggestion
		return pullImage(ctx, client, &buildList.Items[0])
```

internal/format/print.go
```diff
@@ -78,3 +87,116 @@ func spinnerConfig(message, stopMessage string) yacspin.Config {
...
+	annotations := strippedObj.GetAnnotations()
+	for k := range annotations {
+		if strings.HasPrefix(k, "crossplane.io") ||
+			strings.HasPrefix(k, "kubectl.kubernetes.io") {
+			delete(annotations, k)
+		}
+	}
+	strippedObj.SetAnnotations(annotations)
+
+	// some fields cannot be removed with a Set, so we convert to unstructured and
```
thirdeyenick, line 185:
```suggestion
	// some fields cannot be removed with a Set, so we convert to unstructured
```

thirdeyenick (approved):

PR #20, opened 2023-05-31
Add app update command

ctrox:
This adds an update command for the app. While it is possible to update an app using apply, it's not exactly the most user-friendly and also not great for scripting.

### DIFF
--- a/api/util/apps.go
+++ b/api/util/apps.go
@@ -1,8 +1,17 @@
 package util
 
-import apps "github.com/ninech/apis/apps/v1alpha1"
+import (
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
 
-const ApplicationNameLabel = "application.apps.nine.ch/name"
+const (
+	ApplicationNameLabel = "application.apps.nine.ch/name"
+	PrivateKeySecretKey  = "privatekey"
+	UsernameSecretKey    = "username"
+	PasswordSecretKey    = "password"
+)
 
 func UnverifiedAppHosts(app *apps.Application) []string {
 	unverifiedHosts := []string{}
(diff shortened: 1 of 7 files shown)


update/application.go
```diff
@@ -0,0 +1,105 @@
...
+// all fields need to be pointers so we can detect if they have been set by
+// the user.
+type applicationCmd struct {
+	Name     *string            `arg:"" help:"Name of the application."`
+	Git      *gitConfig         `embed:"" prefix:"git-"`
+	Size     *string            `help:"Size of the app."`
+	Port     *int32             `help:"Port the app is listening on."`
+	Replicas *int32             `help:"Amount of replicas of the running app."`
+	Hosts    *[]string          `help:"Host names where the application can be accessed. If empty, the application will just be accessible on a generated host name on the deploio.app domain."`
+	Env      *map[string]string `help:"Environment variables which are passed to the app at runtime."`
```
pawelkuc, line 22: we do not need pointer for slices and maps, right? 

> Map types are reference types, like pointers or slices...

https://go.dev/blog/maps#declaration-and-initialization

update/application.go
```diff
@@ -0,0 +1,105 @@
...
+
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/api/util"
+	"github.com/ninech/nctl/internal/format"
+)
+
+// all fields need to be pointers so we can detect if they have been set by
+// the user.
+type applicationCmd struct {
```
pawelkuc, line 15: we are basically copying all tags form create/application.go. I'm just wondering if we can reuse it and somehow define them all in one place. Otherwise, it will be a bit difficult to maintain it.

The same for type `gitConfig`

update/application_test.go
```diff
@@ -0,0 +1,188 @@
...
+			},
+			cmd: applicationCmd{
+				Name: pointer.String(existingApp.Name),
+				Git: &gitConfig{
+					Username: pointer.String("new-user"),
+					Password: pointer.String("new-pass"),
+				},
+			},
+			checkSecret: func(t *testing.T, cmd applicationCmd, authSecret *corev1.Secret) {
+				assert.Equal(t, *cmd.Git.Username, string(authSecret.Data["username"]))
```
pawelkuc, line 117: missing "password" check

update/application.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := client.Update(ctx, app); err != nil {
+		return err
+	}
+
+	format.PrintSuccessf("⬆️", "updated %s %q", apps.ApplicationKind, app.Name)
+
+	return nil
+}
+
+func (cmd *applicationCmd) applyUpdates(app *apps.Application) {
```
pawelkuc, line 77: for application create we have create/create.go with type `creator` + methods. Maybe we should do the same for update? I mean create type `updater` + methods. Just an idea 🙂

pawelkuc (approved):

update/application.go
ctrox: But what if you want to empty the env with an update call? We have to able to tell if the user specified the flag `env=""` (meaning the update call should send that to the server) or the flag was not specified (Env == nil), meaning we should not touch the env.

update/application.go
ctrox: I usually added these as soon as we had 2+ resources, that usually makes it clearer what exactly the helper will need to do. But yeah for updates it should always be roughly the same steps so I'll add something 👍

api/util/apps.go
```diff
@@ -23,3 +27,62 @@ func VerifiedAppHosts(app *apps.Application) []string {
...
+		},
+		Data: data,
+	}
+}
+
+// Update replaces the data of the secret with the data from GitAuth. Only
+// replaces fields which are non-nil.
+func (git GitAuth) Update(secret *corev1.Secret) {
+	if git.SSHPrivateKey != nil {
+		secret.Data["privatekey"] = []byte(*git.SSHPrivateKey)
```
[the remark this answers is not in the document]
ctrox: I added consts for all of these

api/util/apps.go
```diff
@@ -23,3 +27,62 @@ func VerifiedAppHosts(app *apps.Application) []string {
...
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: data,
+	}
+}
+
+// Update replaces the data of the secret with the data from GitAuth. Only
+// replaces fields which are non-nil.
+func (git GitAuth) Update(secret *corev1.Secret) {
```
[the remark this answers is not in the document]
ctrox: Agreed, I renamed it to `UpdateSecret`, does that make more sense?

update/application.go
ctrox: I added an `updater` helper. It's quite basic:

* Get the current resource from the API
* Call the updateFunc where all the custom logic per resource happens
* Call update on the API
* Print success message

update/application.go
ctrox: Sadly that is not how kong behaves as it will initialize the map for you. Given these flags:
```golang
Env      map[string]string
EnvPtr   *map[string]string

// cmd code
fmt.Println(cmd.Env)
fmt.Println(cmd.EnvPtr)
```

```
# without setting any flag
$ go run .
map[]
<nil>

# same when setting --env
$ go run . --env=""
map[]
<nil>

# when setting --env-ptr
$ go run . --env-ptr=""
map[]
&map[]
```

pawelkuc: then my bad. Sorry. I did not know that. This is a little different behavior than what you'd expect from Go.

Do you think that we could cover this case (empty Env vs nil) in our tests? Or Is it too much?

PR #21, opened 2023-06-01
Output build/release logs on failure

ctrox:


create/application.go
```diff
@@ -286,3 +293,34 @@ func printUnverifiedHostsMessage(app *apps.Application) {
...
+		errorLogQuery(logs.ApplicationQuery(release.Labels[util.ApplicationNameLabel], release.Namespace)),
+	)
+}
+
+const errorLogLines = 20
+
+func errorLogQuery(queryString string) log.Query {
+	// we print the last 20 lines of the log. In most cases this should be
+	// enough to give a hint about the problem but we might need to tweak this
+	// value a bit in the future.
```
pawelkuc, line 316: I would move this comment up, above `const errorLogLines = 20` line

internal/test/client.go
```diff
@@ -0,0 +1,20 @@
...
+func SetupClient(initObjs ...client.Object) (*api.Client, error) {
+	scheme, err := api.NewScheme()
+	if err != nil {
+		return nil, err
+	}
+
+	client := fake.NewClientBuilder().WithScheme(scheme).WithObjects(initObjs...).Build()
+
+	return &api.Client{
+		WithWatch: client, Namespace: "default",
```
pawelkuc, line 18: I like the idea to have a common logic for test setup!

one suggestion: maybe we should expose the namespace argument as well? This will make it more universal. Additionally, having this hardcoded we need to remember to use the "default" namespace inside the test - it's a bit error prone.

I will also change my PR to match this new SetupClient() func.

create/application_test.go
```diff
@@ -238,6 +236,92 @@ func TestApplicationWait(t *testing.T) {
...
+				if err := client.Status().Update(ctx, build); err != nil {
+					errors <- err
+				}
+			}
+		}
+	}()
+
+	if err := cmd.Run(ctx, client); err == nil {
+		t.Fatal("expected build error, got nil")
+	}
```
pawelkuc, line 312: maybe that is just me but I do not see these last 20 lines of the log after executing this block. My output:
```bash
Creating a new application
 ✓ created application "some-name" 🏗
 ⠋ waiting for build to start ⏳
 ✗ waiting for build to start ⏳

Your build has failed with status "error". Here are the last 20 lines of the log:
```
(zero log lines)

logs/logs_test.go
```diff
@@ -79,7 +79,7 @@ func TestApplication(t *testing.T) {
 
 			tc.cmd.out = out
 
-			if err := tc.cmd.Run(ctx, apiClient, "app", "app-name"); err != nil {
+			if err := tc.cmd.Run(ctx, apiClient, ApplicationQuery("app-name", "app-ns")); err != nil {
```
pawelkuc, line 82: missing similar test for `BuildQuery`

pawelkuc (approved):

internal/test/client.go
ctrox: It's already exposed as it's an exported field on the client ;) I kind of don't want to set it all the time in an argument when it's going to be `default` in almost all cases.

create/application_test.go
ctrox: Yeah that is to be expected in the test output as we redirect the logs to a `bytes.Buffer` so we can do some checks on it.

logs/logs_test.go
ctrox: You are right but I felt like it's sort of pointless as the only thing that changes is the query string. Since we don't have a real loki the query doesn't really do anything.

create/application_test.go
pawelkuc: true 🤦

PR #22, opened 2023-06-01
Add release history

pawelkuc:
This adds `get releases` command (view your app’s release history).

### DIFF
--- a/get/get.go
+++ b/get/get.go
@@ -18,6 +18,7 @@ type Cmd struct {
 	APIServiceAccounts apiServiceAccountsCmd `cmd:"" group:"iam.nine.ch" name:"apiserviceaccounts" aliases:"asa" help:"Get API Service Accounts."`
 	Applications       applicationsCmd       `cmd:"" group:"deplo.io" name:"applications" aliases:"app,apps" help:"Get deplo.io Applications. (Beta - requires access)"`
 	Builds             buildCmd              `cmd:"" group:"deplo.io" name:"builds" aliases:"build" help:"Get deplo.io Builds. (Beta - requires access)"`
+	Releases           releasesCmd           `cmd:"" group:"deplo.io" name:"releases" aliases:"release" help:"Get deplo.io Releases. (Beta - requires access)"`
 
 	opts []runtimeclient.ListOption
 }
--- a/get/releases.go
+++ b/get/releases.go
@@ -1,0 +1,120 @@
+package get
+
+import (
+	"context"
+	"io"
+	"os"
+	"sort"
+	"strconv"
+	"text/tabwriter"
+	"time"
+
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/api/util"
+	"github.com/ninech/nctl/internal/format"
+	"k8s.io/apimachinery/pkg/util/duration"
+)
+
+type releasesCmd struct {
+	Name            string `arg:"" help:"Name of the Release to get. If omitted all in the namespace will be listed." default:""`
+	ApplicationName string `help:"Name of the Application to get releases for. If omitted all in the namespace will be listed."`
+	out             io.Writer
+}
+
+func (cmd *releasesCmd) Run(ctx context.Context, client *api.Client, get *Cmd) error {
+	releaseList := &apps.ReleaseList{}
+
+	opts := []listOpt{matchName(cmd.Name)}
+	if len(cmd.ApplicationName) != 0 {
+		opts = append(opts, matchLabel(util.ApplicationNameLabel, cmd.ApplicationName))
+	}
+
+	if err := get.list(ctx, client, releaseList, opts...); err != nil {
+		return err
+	}
+
+	if len(releaseList.Items) == 0 {
+		printEmptyMessage(apps.ReleaseKind, client.Namespace)
+		return nil
+	}
+
+	switch get.Output {
+	case full:
+		return printReleases(releaseList.Items, get, true)
+	case noHeader:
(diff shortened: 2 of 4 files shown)


get/get.go
```diff
@@ -18,6 +19,7 @@ type Cmd struct {
 	APIServiceAccounts apiServiceAccountsCmd `cmd:"" group:"iam.nine.ch" name:"apiserviceaccounts" aliases:"asa" help:"Get API Service Accounts."`
 	Applications       applicationsCmd       `cmd:"" group:"deplo.io" name:"applications" aliases:"app,apps" help:"Get deplo.io Applications. (Beta - requires access)"`
 	Builds             buildCmd              `cmd:"" group:"deplo.io" name:"builds" aliases:"build" help:"Get deplo.io Builds. (Beta - requires access)"`
+	Releases           releasesCmd           `cmd:"" group:"deplo.io" name:"releases" help:"Get deplo.io Releases. (Beta - requires access)"`
```
ctrox, line 22: Can we also add an alias for the singular `release`?

get/releases.go
```diff
@@ -0,0 +1,94 @@
...
+			get.writeTabRow(
+				w,
+				r.ObjectMeta.Namespace,
+				r.Spec.ForProvider.Build.Name,
+				r.ObjectMeta.Name,
+				r.ObjectMeta.Labels[util.ApplicationNameLabel],
+				string(*r.Spec.ForProvider.Config.Size),
+				strconv.Itoa(int(*r.Spec.ForProvider.Config.Replicas)),
+				string(r.Status.AtProvider.ReleaseStatus),
+				r.ObjectMeta.CreationTimestamp.String(),
```
ctrox, line 88: Instead of printing the whole timestamp, can we display the relative age of the release? I already did this in the build using `duration.HumanDuration(time.Since(build.CreationTimestamp.Time)))`. It's the same function `kubectl` calls to populate the age column.

get/releases.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, orderedRelease := range orderedReleases {
+		for _, r := range orderedRelease {
+			get.writeTabRow(
+				w,
+				r.ObjectMeta.Namespace,
+				r.Spec.ForProvider.Build.Name,
+				r.ObjectMeta.Name,
+				r.ObjectMeta.Labels[util.ApplicationNameLabel],
+				string(*r.Spec.ForProvider.Config.Size),
+				strconv.Itoa(int(*r.Spec.ForProvider.Config.Replicas)),
```
ctrox, line 86: Nit: Potential nil pointers. While these fields should never be empty by the time a release is created, we should probably still check it.

get/releases.go
```diff
@@ -0,0 +1,94 @@
...
+	return nil
+}
+
+// prepareReleasesOutput prepares Release collection for printing
+func prepareReleasesOutput(releases []apps.Release) map[string][]apps.Release {
+	orderedReleases := make(map[string][]apps.Release)
+
+	for _, release := range releases {
+		orderedReleases[release.ObjectMeta.Labels[util.ApplicationNameLabel]] =
+			append(orderedReleases[release.ObjectMeta.Labels[util.ApplicationNameLabel]], release)
```
ctrox, line 47: Nit: assignment over two lines looks a bit strange, maybe put `release.ObjectMeta.Labels[util.ApplicationNameLabel]` in a variable so this line can be shorter?

get/releases.go
```diff
@@ -0,0 +1,94 @@
...
+	if len(releaseList.Items) == 0 {
+		printEmptyMessage(apps.ReleaseKind, client.Namespace)
+		return nil
+	}
+
+	switch get.Output {
+	case full:
+		return printReleases(releaseList.Items, get, true)
+	case noHeader:
+		return printReleases(releaseList.Items, get, false)
```
ctrox, line 35: Can you also add yaml output?

```golang
case yamlOut:
	return format.PrettyPrintObjects(releaseList.GetItems(), format.PrintOpts{})
```

get/releases_test.go
```diff
@@ -0,0 +1,248 @@
...
+		Replicas: pointer.Int32(1),
+		Port:     pointer.Int32(8080),
+	}
+
+	defaultCreationTime = metav1.NewTime(mustParseTime(time.RFC3339, "2023-03-13T14:00:00Z"))
+)
+
+func mustParseTime(format string, value string) time.Time {
+	t, _ := time.Parse(format, value)
+	return t
```
ctrox, line 41: Shouldn't `must*` functions usually panic when running into an error? This is more of a `maybeParseTime` 😄

get/releases.go
```diff
@@ -0,0 +1,94 @@
...
+
+func printReleases(releases []apps.Release, get *Cmd, header bool) error {
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 4, ' ', 0)
+
+	if header {
+		get.writeHeader(
+			w,
+			"BUILDNAME",
+			"NAME",
+			"APPLICATION",
```
ctrox, line 67: I feel like the first column should always be the name of the current resource. Buildname second or third does not matter that much IMO.

get/releases.go
```diff
@@ -0,0 +1,94 @@
...
+	"strconv"
+	"text/tabwriter"
+
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/api/util"
+)
+
+type releasesCmd struct {
+	Name string `arg:"" help:"Name of the Application for which to get the Releases. If omitted all Applications will be listed." default:""`
```
ctrox, line 16: Can we also add an optional flag to get releases by application name? See `Build` for an example.

ctrox (approved):

PR #24, opened 2023-06-09
Add config create/delete/get/update commands

pawelkuc:
This adds `create/delete/get/update config` commands to enable CRUD of ProjectConfiguration resource.

### DIFF
--- a/api/util/apps.go
+++ b/api/util/apps.go
@@ -1,6 +1,9 @@
 package util
 
 import (
+	"fmt"
+	"strings"
+
 	apps "github.com/ninech/apis/apps/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -39,6 +42,19 @@ func EnvVarsFromMap(env map[string]string) apps.EnvVars {
 		vars = append(vars, apps.EnvVar{Name: k, Value: v})
 	}
 	return vars
+}
+
+func EnvVarToString(envs apps.EnvVars) string {
+	if envs == nil {
+		return "nil"
+	}
+
+	var keyValuePairs []string
+	for _, env := range envs {
+		keyValuePairs = append(keyValuePairs, fmt.Sprintf("%v=%v", env.Name, env.Value))
+	}
+
+	return strings.Join(keyValuePairs, ";")
 }
 
 type GitAuth struct {
--- a/create/create.go
+++ b/create/create.go
@@ -23,6 +23,7 @@ type Cmd struct {
 	VCluster          vclusterCmd          `cmd:"" group:"infrastructure.nine.ch" name:"vcluster" help:"Create a new vcluster."`
 	APIServiceAccount apiServiceAccountCmd `cmd:"" group:"iam.nine.ch" name:"apiserviceaccount" aliases:"asa" help:"Create a new API Service Account."`
 	Application       applicationCmd       `cmd:"" group:"deplo.io" name:"application" aliases:"app" help:"Create a new deplo.io Application. (Beta - requires access)"`
+	Config            configCmd            `cmd:"" group:"deplo.io" name:"config"  help:"Create a new deplo.io Project Configuration. (Beta - requires access)"`
 }
 
 // resultFunc is the function called on a watch event during creation. It
--- a/create/project_config.go
+++ b/create/project_config.go
@@ -1,0 +1,64 @@
+package create
+
+import (
+	"context"
+
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/api/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// all fields need to be pointers so we can detect if they have been set by
+// the user.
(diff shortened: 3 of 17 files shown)


create/project_config.go
```diff
@@ -0,0 +1,61 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+type configCmd struct {
+	Wait        bool              `default:"true" help:"Wait until API Project Config is fully created."`
+	WaitTimeout time.Duration     `default:"10s" help:"Duration to wait for API Project Config getting ready. Only relevant if wait is set."`
+	Size        string            `default:"micro" help:"Size of the app."`
+	Port        int32             `default:"8080" help:"Port the app is listening on."`
+	Replicas    int32             `default:"1" help:"Amount of replicas of the running app."`
```
ctrox, line 19: Not sure about these defaults, seems very dangerous. If I do `nctl create config --replicas=3`, all my apps will suddenly listen on port `8080` (unless the app overrides it) even though I never intended to configure that. IMO just the args defined by the user should be set in the config and the rest should be `nil`.

create/project_config.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	if !cmd.Wait {
+		return nil
+	}
+
+	return c.wait(ctx, waitStage{
+		objectList: &apps.ProjectConfigList{},
+		onResult:   resourceAvailable,
```
ctrox, line 38: I think this won't work for `ProjectConfigs` since we have no controller setting these resources to available. Essentially as soon as they are created they should be considered available as a new app will make use of them immediately.

create/project_config_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/ninech/apis/apps/v1alpha1"
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/api/util"
+	"github.com/ninech/nctl/internal/test"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProjectConfig(t *testing.T) {
+
```
ctrox, line 16: Nit: whitespace

create/create.go
```diff
@@ -23,6 +23,7 @@ type Cmd struct {
 	VCluster          vclusterCmd          `cmd:"" group:"infrastructure.nine.ch" name:"vcluster" help:"Create a new vcluster."`
 	APIServiceAccount apiServiceAccountCmd `cmd:"" group:"iam.nine.ch" name:"apiserviceaccount" aliases:"asa" help:"Create a new API Service Account."`
 	Application       applicationCmd       `cmd:"" group:"deplo.io" name:"application" aliases:"app" help:"Create a new deplo.io Application. (Beta - requires access)"`
+	Config            configCmd            `cmd:"" group:"deplo.io" name:"config"  help:"Create a new deplo.io Project Configuration. (Beta - requires access)"`
```
ctrox, line 26: Nit: I noticed some slight naming inconsistencies, sometimes we call it `Project Config` and other times `Project Configuration`.

api/util/apps.go
```diff
@@ -41,6 +44,19 @@ func EnvVarsFromMap(env map[string]string) apps.EnvVars {
...
 }
 
+func EnvVarToString(envs apps.EnvVars) string {
+	if envs == nil {
+		return "nil"
+	}
+
+	var s strings.Builder
+	for _, env := range envs {
+		s.WriteString(fmt.Sprintf("%v=%v;", env.Name, env.Value))
```
ctrox, line 54: A bit of a nit: If there is just one env it will be `key=val;`, which looks a bit off. I guess we could construct a `[]string` with just `key=val` for each env and then do a return `strings.Join([]string, ";")` at the end.

get/project_config.go
```diff
@@ -0,0 +1,116 @@
...
+			"SIZE",
+			"REPLICAS",
+			"PORT",
+			"ENVIRONMENT_VARIABLES",
+			"AGE",
+		)
+	}
+
+	for _, c := range configs {
+
```
ctrox, line 85: Nit: whitespace

get/project_config.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return nil
+}
+
+// filterProjectConfigList filters out everything except the ProjectConfig
+// resource whose Name is equal to its Namespace (as defined in the project
+// documentation, a valid Project Configuration has Name=Namespace).
+// If no such resource is found, func returns an empty slice.
+func filterProjectConfigList(configList *apps.ProjectConfigList) {
```
ctrox, line 57: Do we really need this? We will have a webhook in place not allowing to create any project configs like that anyways.

create/project_config.go
pawelkuc: I completely agree

create/project_config.go
pawelkuc: true 🤦

create/project_config.go
ctrox: The defaults are still there, I guess we need to remove them else kong will just set the flags.

pawelkuc: my bad - sorry.

ctrox (approved):

PR #28, opened 2023-06-16
fix: Improve wait error handling

ctrox:
Whenever a watch error happens, we used to show a line like "watch failed". This can be very confusing to the user as we anyways retry when that happens and it has nothing to do with the actual creation process of the resource. This hides the error completely and unless we are not running into the backoff it will just continue. The backoff also has been increased to account for the API cancelling watches sometimes.


thirdeyenick (approved):
looks good!

PR #31, opened 2023-06-20
Tag pulled build images with latest

ctrox:
Without a tag, using the pulled image with docker/podman is a bit cumbersome as you have to first find out the image ID. With a tag it's possible to just do `docker run <image name>`.

### DIFF
--- a/get/build.go
+++ b/get/build.go
@@ -113,9 +113,29 @@ func pullImage(ctx context.Context, apiClient *api.Client, build *apps.Build) er
 		return err
 	}
 
+	if err := tagImage(ctx, cli, build); err != nil {
+		return fmt.Errorf("unable to tag image: %w", err)
+	}
+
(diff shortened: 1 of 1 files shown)


gajicdev:
should we maybe add a way of passing a "custom" tag?

ctrox:
> should we maybe add a way of passing a "custom" tag?

Not sure, you can still that afterwards using `docker tag`. But now that you mention it, it would make sense to add a tag with the build name in addition of the `latest` tag. So if you pull multiple versions you can easily tell them apart by the build name but the latest pulled will be the latest, making it easy to do a `docker run` without being specific about the tag.

gajicdev:
yeah that would be great. Just in case you want to compare the versions locally 👍

gajicdev (approved):
looks good 👍

PR #34, opened 2023-06-20
delete secrets created by nctl when deleting applications

thirdeyenick:
This will delete git auth secrets which were created by `nctl` on application delete. If the secret is still referenced by another application, a corresponding message will be printed.

### DIFF
--- a/api/util/apps.go
+++ b/api/util/apps.go
@@ -11,6 +11,8 @@ import (
 
 const (
 	ApplicationNameLabel = "application.apps.nine.ch/name"
+	ManagedByAnnotation  = "app.kubernetes.io/managed-by"
+	NctlName             = "nctl"
 	PrivateKeySecretKey  = "privatekey"
 	UsernameSecretKey    = "username"
 	PasswordSecretKey    = "password"
@@ -63,7 +65,7 @@ type GitAuth struct {
 	SSHPrivateKey *string
 }
 
-func (git GitAuth) Secret(name, project string) *corev1.Secret {
+func (git GitAuth) Secret(app *apps.Application) *corev1.Secret {
 	data := map[string][]byte{}
 
 	if git.SSHPrivateKey != nil {
@@ -75,8 +77,11 @@ func (git GitAuth) Secret(name, project string) *corev1.Secret {
 
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: project,
+			Name:      GitAuthSecretName(app),
+			Namespace: app.Namespace,
+			Annotations: map[string]string{
+				ManagedByAnnotation: NctlName,
+			},
 		},
 		Data: data,
 	}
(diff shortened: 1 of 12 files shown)


api/util/apps.go
```diff
@@ -11,6 +11,8 @@ import (
 
 const (
 	ApplicationNameLabel = "application.apps.nine.ch/name"
+	ManagedByAnnotation  = "app.kubernetes.io/managed-by"
+	NctlName             = "nctl"
```
ctrox, line 15: We should probably have this const somewhere we can use it from all packages? Grepping the project I already found two existing instances of the command name `"nctl"`:

```
main.go
54:             kong.Name("nctl"),

auth/login.go
28:     NctlExtensionName = "nctl"
```

Not yet sure where it fits best though.

delete/application.go
```diff
@@ -27,12 +32,117 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
+			return nil
+		}
+	}
+	if err := client.Delete(ctx, &secret); err != nil {
+		return checkManuallyError(
+			fmt.Errorf(
+				"error when deleting git auth secret %q: %w",
+				secret.Name,
+				err,
+			),
```
ctrox, line 143: Nit: This could probably fit on one line just fine?

delete/application_test.go
```diff
@@ -5,41 +5,198 @@ import (
...
+					appOne,
+					nctlSecret,
+					noDeletionExpected(customSecret),
+				)
+			}(),
+			name:          "dev",
+			errorExpected: false,
+		},
+		// a secret which has the same name as the application, but no
+		// nctl annotation will no be deleted
```
ctrox, line 94:
```suggestion
		// nctl annotation will not be deleted
```

delete/application_test.go
```diff
@@ -5,41 +5,198 @@ import (
...
+					appOne,
+					noDeletionExpected(appTwo),
+					noDeletionExpected(secret),
+				)
+			}(),
+			name:          "dev",
+			errorExpected: false,
+		},
+		// here we have 2 secrets. One secret created by nctl which is
+		// not in use and another secret non nctl managed secret which
```
ctrox, line 69:
```suggestion
		// not in use and another non nctl managed secret which
```

api/util/apps.go
thirdeyenick: I put it into the `util` package, but I am open for any other location

delete/application.go
```diff
@@ -27,12 +32,117 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
+			return nil
+		}
+		return checkManuallyError(fmt.Errorf("error when checking git auth secret %q for application", secret.Name))
+	}
+	managedBy, exists := secret.Annotations[util.ManagedByAnnotation]
+	if !exists || managedBy != util.NctlName {
+		fmt.Printf("will not delete git auth secret %q as it does not contain expected annotation %q",
+			secret.Name,
+			fmt.Sprintf("%s: %s", util.ManagedByAnnotation, util.NctlName),
+		)
```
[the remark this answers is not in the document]
thirdeyenick: I can also leave it out. The 2 mentioned cases ("secret with git auth creds made by user" and "secret with other use case made by user") are hard to differentiate. I have no strong preference here.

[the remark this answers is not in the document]
thirdeyenick: I removed it for now.

ctrox (approved):

PR #35, opened 2023-06-21
find absolute path to nctl binary

thirdeyenick:
This will resolve the absolute path of the nctl binary so it can be used as credential helper in the kubeconfig. It allows for locally build "nctl" versions and calling them via "./nctl" for example.

### DIFF
--- a/auth/login.go
+++ b/auth/login.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"os/exec"
 
 	"github.com/ninech/nctl/api"
 	"github.com/ninech/nctl/api/util"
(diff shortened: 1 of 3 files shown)


auth/login.go
```diff
@@ -132,11 +131,6 @@ func newAPIConfig(apiURL, issuerURL *url.URL, command, clientID string, opts ...
 		return clientConfig, nil
 	}
 
-	// we make sure our command is in the $PATH as the client-go credential plugin will need to find it.
-	if _, err := exec.LookPath(command); err != nil && command != "" {
```
ctrox, line 136: AFAIK this change only applies to the `nctl auth cluster` command, not `nctl auth login`. As `auth login` also adds an entry to the kubeconfig with the command `nctl`, we still need to ensure `nctl` is in the path or we need to also change `auth login`.

thirdeyenick: I changed `auth login` to also use the same mechanism.

ctrox (approved):

PR #40, opened 2023-07-04
Add build-time env vars to application

pawelkuc:
We should support buildpack env variables. This PR adds new flags to support passing these build-time env vars on app creation/update.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/lucasepe/codename v0.2.0
 	github.com/moby/moby v24.0.1+incompatible
 	github.com/moby/term v0.0.0-20210619224110-3f7ff695adc6
-	github.com/ninech/apis v0.0.0-20230630090309-a9de5a5a3b7e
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/lucasepe/codename v0.2.0
 	github.com/moby/moby v24.0.1+incompatible
 	github.com/moby/term v0.0.0-20210619224110-3f7ff695adc6
-	github.com/ninech/apis v0.0.0-20230630090309-a9de5a5a3b7e
+	github.com/ninech/apis v0.0.0-20230704182350-ac487125446e
```
pawelkuc, line 20: Using a temporary [branch](https://github.com/ninech/apis/tree/add-build-env) here. This will be changed to main once github.com/ninech/apis is synced with the latest version.

pawelkuc: Done. Using the latest main now.

thirdeyenick (approved):

PR #43, opened 2023-07-11
do not set defaults in the application on creation

thirdeyenick:
If we set defaults in the application on creation, existing content in the .deploio.yaml file will be ignored (as the config which is set on the Application itself has higher precedence). As we will always fall back to valid defaults, we can just leave them out in the application if the user did not set an explicit value.

### DIFF
--- a/create/application.go
+++ b/create/application.go
@@ -27,23 +27,23 @@ import (
 // note: when adding/changing fields here also make sure to carry it over to
 // update/application.go.
 type applicationCmd struct {
-	Name        string            `arg:"" default:"" help:"Name of the application. A random name is generated if omitted."`
-	Wait        bool              `default:"true" help:"Wait until application is fully created."`
-	WaitTimeout time.Duration     `default:"15m" help:"Duration to wait for application getting ready. Only relevant if wait is set."`
+	Name        string            `arg:"" default:"" help:"Name of the app. A random name is generated if omitted."`
+	Wait        bool              `default:"true" help:"Wait until the app is fully created."`
+	WaitTimeout time.Duration     `default:"15m" help:"Duration to wait for the app getting ready. Only relevant if wait is set."`
 	Git         gitConfig         `embed:"" prefix:"git-"`
-	Size        *string           `help:"Size of the app (defaults to \"${app_default_size}\")."`
-	Port        *int32            `help:"Port the app is listening on (defaults to ${app_default_port})."`
-	Replicas    *int32            `help:"Amount of replicas of the running app (defaults to ${app_default_replicas})."`
-	Hosts       []string          `help:"Host names where the application can be accessed. If empty, the application will just be accessible on a generated host name on the deploio.app domain."`
-	BasicAuth   *bool             `help:"Enable/Disable basic authentication for the applicaton (defaults to ${app_default_basic_auth})."`
+	Size        *string           `help:"Size of the app (defaults to \"${app_default_size}\")." placeholder:"${app_default_size}"`
+	Port        *int32            `help:"Port the app is listening on (defaults to ${app_default_port})." placeholder:"${app_default_port}"`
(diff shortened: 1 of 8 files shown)


create/application.go
```diff
@@ -29,11 +31,11 @@ type applicationCmd struct {
...
 	Git         gitConfig         `embed:"" prefix:"git-"`
-	Size        string            `default:"micro" help:"Size of the app."`
-	Port        int32             `default:"8080" help:"Port the app is listening on."`
-	Replicas    int32             `default:"1" help:"Amount of replicas of the running app."`
+	Size        *string           `help:"Size of the app (defaults to \"${app_default_size}\")."`
+	Port        *int32            `help:"Port the app is listening on (defaults to ${app_default_port})."`
+	Replicas    *int32            `help:"Amount of replicas of the running app (defaults to ${app_default_replicas})."`
 	Hosts       []string          `help:"Host names where the application can be accessed. If empty, the application will just be accessible on a generated host name on the deploio.app domain."`
-	BasicAuth   bool              `default:"false" help:"Enable/Disable basic authentication for the application."`
+	BasicAuth   *bool             `help:"Enable/Disable basic authentication for the applicaton (defaults to ${app_default_basic_auth})."`
```
ctrox, line 38: Nit: some consistency would be nice, maybe call it `app` everywhere in the help instead of `application`?

create/application.go
```diff
@@ -29,11 +31,11 @@ type applicationCmd struct {
 	Wait        bool              `default:"true" help:"Wait until application is fully created."`
 	WaitTimeout time.Duration     `default:"15m" help:"Duration to wait for application getting ready. Only relevant if wait is set."`
 	Git         gitConfig         `embed:"" prefix:"git-"`
-	Size        string            `default:"micro" help:"Size of the app."`
-	Port        int32             `default:"8080" help:"Port the app is listening on."`
-	Replicas    int32             `default:"1" help:"Amount of replicas of the running app."`
+	Size        *string           `help:"Size of the app (defaults to \"${app_default_size}\")."`
```
ctrox, line 34: I just noticed something when looking at the help of this branch. Whenever a flag is a pointer, the help will look like this: `--size=SIZE` instead of `--size=STRING`. This is IMO especially problematic for bools, such as `--basic-auth=BASIC-AUTH`. It's rather confusing to know what I now need to supply here. Should we maybe use the `placeholder` tag for these? So e.g. `placeholder="micro"`, so the help shows a real example like `--size=micro` (which is also like it was with the default before).

But sadly it looks like the placeholder does not support variables.

thirdeyenick: Good catch. I will have a look. As I tested with a different approach yesterday I might have an idea.

create/application.go
ctrox: I don't have a strong preference on either just that `app` is making the help a bit shorter. Other than that consistency is more important.

thirdeyenick: I replaced all occurrences of 'application' with 'app' in the create command.

internal/format/interpolation.go
```diff
@@ -0,0 +1,76 @@
...
+
+// interpolationRegex is the regex to find if a given string contains variables
+// for interpolation
+var interpolationRegex = regexp.MustCompile(`(\$\$)|((?:\${([[:alpha:]_][[:word:]]*))(?:=([^}]+))?})|(\$)|([^$]+)`)
+
+// interpolate interpolates the given string s with variables from vars
+// The [upstream
+// function](https://github.com/alecthomas/kong/blob/v0.8.0/interpolate.go#L22)
+// was sadly not exported, so we had to copy it.
+func interpolate(s string, vars kong.Vars) (string, error) {
```
ctrox, line 19: pretty nasty 😄 I guess the best would be to eventually get this upstream (so placeholders are interpolated automatically) so we don't need to copy this anymore.

PR #44, opened 2023-07-14
Do not error on missing command/argument

pawelkuc:
Do not error on missing command/argument. Print Usage + friendly message instead (and exit). For example:

missing command:
```bash
> nctl
Usage: nctl <command> 
...
Your query: "nctl": expected one of "get",  "auth",  "completions",  "create",  "apply",  ...
```

```bash
> nctl get
Usage: nctl get <command>
...
Your query: "nctl get": expected one of "clusters",  "apiserviceaccounts",  "projects",  "applications",  "builds",  ...
```

missing arg:
```bash
> nctl delete project
Usage: nctl delete project (proj) <name>
...
Your query: "nctl delete project": expected "<name>"
```

### DIFF
--- a/internal/format/command.go
+++ b/internal/format/command.go
@@ -1,9 +1,12 @@
 package format
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
+
+	"github.com/alecthomas/kong"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/format/command.go
```diff
@@ -30,3 +33,59 @@ func (c command) GetApplication(extraFields ...string) string {
...
+
+	query := parseErr.Context.Model.Name
+	if len(args) > 0 {
+		queryArgs := fmt.Sprintf(args[0].(string), args[1:]...)
+		if len(queryArgs) > 0 {
+			query += " " + queryArgs
+		}
+	}
+
+	fmt.Printf("\n💡 Your query: %q: %s\n", query, msg)
```
ctrox, line 86: Not a 100% about the word `query` here. Wouldn't `command` be more accurate?

internal/format/command.go
```diff
@@ -30,3 +33,59 @@ func (c command) GetApplication(extraFields ...string) string {
...
+		}
+
+		return true
+	}
+
+	return false
+}
+
+// NoFatalExitIfErrorf prints Usage + friendly message on error (and exits).
+func NoFatalExitIfErrorf(err error, args ...interface{}) error {
```
ctrox, line 64: This function name is IMO a bit misleading as it's still a fatal exit (exit code 1).

main.go
```diff
@@ -71,6 +72,23 @@ func main() {
...
+					node = parseErr.Context.Model.Node
+				}
+				if format.MissingChildren(node) {
+					err = format.NoFatalExitIfErrorf(err, parseErr.Context.Command())
+				}
+			}
+		}
+	}
+
 	parser.FatalIfErrorf(err)
```
ctrox, line 92: Nit: this could be moved up into the `if err != nil` block or maybe even removed? Is there any case where this is still needed?

internal/format/command.go
pawelkuc: true. changing this into `ExitIfErrorf()`

main.go
pawelkuc: moving this up as suggested.

yes we need it - we still error on unknown argument, e.g.:
```bash
> nctl dummy
...
nctl: error: unexpected argument dummy
```

ctrox (approved):

PR #46, opened 2023-07-18
clean up auth flow, implement whoami and set-org

gajicdev:


api/login.go
```diff
@@ -159,3 +172,38 @@ func GetToken(ctx context.Context, issuerURL, clientID string, usePKCE bool, out
 
 	return nil
 }
+
+type UserInfo struct {
+	Email string
+	Orgs  []string
+}
+
+func GetUserInfoFromToken(tokenString string) (*UserInfo, error) {
```
ctrox, line 181: Looks like this will probably break for service accounts. It's already possible to login with a service account token using the flag `nctl auth login --api-token`. It's a bit harder to figure out the org from a service account token as it could live in a project namespace (we'd have to make an API call for sure). But maybe in the meantime we can force the user to supply the org also via a flag when `--api-token` is given.

auth/whoami.go
```diff
@@ -0,0 +1,96 @@
...
+
+	userInfo, err := api.GetUserInfoFromToken(client.Token)
+	if err != nil {
+		return err
+	}
+
+	return printUserInfo(userInfo, cfg)
+}
+
+func printUserInfo(userInfo *api.UserInfo, cfg *Config) error {
```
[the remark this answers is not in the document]
gajicdev: switched to just printing lines and cleaned up the whitespaces as suggested

auth/set_org.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return err
+	}
+
+	cfg, err := newAPIConfig(apiURL, issuerURL, command, l.ClientID, withOrganization(l.Organization))
+	if err != nil {
+		return err
+	}
+
+	return login(ctx, cfg, loadingRules.GetDefaultFilename(), runExecPlugin(l.ExecPlugin), project(l.Organization))
```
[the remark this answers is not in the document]
gajicdev: added a line when switching orgs

api/login.go
gajicdev: switch the order and added a org parameter when providing an api token

auth/login.go
```diff
@@ -15,42 +15,71 @@ import (
...
+		return err
+	}
+
+	if len(userInfo.Orgs) == 0 {
+		return fmt.Errorf("error getting an organization for the account %q. Please contact support.", userInfo.User)
 	}
 
-	cfg, err := newAPIConfig(apiURL, issuerURL, command, l.ClientID, opts...)
+	org := userInfo.Orgs[0]
+	if len(userInfo.Orgs) < 1 {
```
ctrox, line 71: Shouldn't this be `> 1`? :)

auth/whoami.go
```diff
@@ -0,0 +1,58 @@
...
+func printUserInfo(userInfo *api.UserInfo, cfg *Config) {
+	fmt.Printf("You are currenlty logged in the with the following account: %q\n", userInfo.User)
+
+	fmt.Printf("Your current organization: %q\n", cfg.Organization)
+
+	printAvailableOrgsString(cfg.Organization, userInfo.Orgs)
+}
+
+func printAvailableOrgsString(currentorg string, orgs []string) {
+	fmt.Print("Active\tOrganization\n")
```
ctrox, line 45: Last nits about this:

* this line should not be printed if there are no orgs, e.g. for a service account
* another line break before would be good to better separate this table from the rest of the output

ctrox (approved):

PR #49, opened 2023-08-18
implement a "get all" command

thirdeyenick:
The "get all" command will show all resources from nine API groups of either a specific project or in all projects.  It won't show nine owned resources by default, but there is a option which allows to list nine owned resources as well.


get/all.go
```diff
@@ -0,0 +1,173 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type allCmd struct {
+	out                  io.Writer
+	stdErr               io.Writer
+	IncludeNineResources bool `arg:"" help:"show resources which are owned by Nine" default:"false"`
```
ctrox, line 26: Why is this an arg? Should be a flag I think.

get/all.go
```diff
@@ -0,0 +1,173 @@
...
+	return result
+}
+
+func getProjectContent(ctx context.Context, client *api.Client, projNames []string, includeNineOwned bool) ([]*unstructured.Unstructured, []string, error) {
+	var warnings []string
+	var result []*unstructured.Unstructured
+	scheme := runtime.NewScheme()
+	if err := apis.AddToScheme(scheme); err != nil {
+		return nil, nil, err
+	}
```
ctrox, line 85: I guess we could also use the existing `client.Scheme()` as it already contains the all `apis` schemes and we filter by nine types anyways.

thirdeyenick: changed this to use the existing client.Scheme

ctrox (approved):

PR #50, opened 2023-08-21
error out when credentials are empty

gajicdev:
Related to https://gitlab.nine.ch/gcp/mono-infra/-/issues/769

### DIFF
--- a/api/util/apps.go
+++ b/api/util/apps.go
@@ -117,6 +117,22 @@ func (git GitAuth) Enabled() bool {
 	return false
 }
 
+func (git GitAuth) Valid() error {
+	if git.SSHPrivateKey != nil {
+		if *git.SSHPrivateKey == "" {
+			return fmt.Errorf("the SSH private key cannot be empty")
+		}
+	}
+
+	if git.Username != nil && git.Password != nil {
+		if *git.Username == "" || *git.Password == "" {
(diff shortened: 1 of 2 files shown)


api/util/apps.go
```diff
@@ -117,6 +117,22 @@ func (git GitAuth) Enabled() bool {
 	return false
 }
 
+func (git GitAuth) Valid() bool {
```
thirdeyenick, line 120: I think we could actually return either a `(bool,error)` or just an `error` to indicate which part has been empty?

gajicdev: changed it to an error 👍

api/util/apps.go
```diff
@@ -117,6 +117,22 @@ func (git GitAuth) Enabled() bool {
...
 
+func (git GitAuth) Valid() error {
+	if git.SSHPrivateKey != nil {
+		if *git.SSHPrivateKey == "" {
+			return fmt.Errorf("the SSH private key cannot be empty")
+		}
+	}
+
+	if git.Username != nil && git.Password != nil {
+		if *git.Username == "" && *git.Password == "" {
```
thirdeyenick, line 128: I think a logical "or" might be better here as both, the username or the password, should not be empty?

gajicdev: yup you are right, added it

thirdeyenick (approved):
👍

PR #51, opened 2023-08-23
add AUR as a goreleaser target

gajicdev:
Closes https://gitlab.nine.ch/gcp/mono-infra/-/issues/752

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -63,6 +63,25 @@ brews:
     # Default is false.
     skip_upload: false
 
+# .goreleaser.yaml
+aurs:
+  - name: nctl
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -63,6 +63,26 @@ brews:
     # Default is false.
     skip_upload: false
 
+# .goreleaser.yaml
+aurs:
+  -
+    name: nctl
```
thirdeyenick, line 69: maybe put that line one line above so that it matches the `-`?

gajicdev: done 👍

thirdeyenick (approved):
👍

PR #58, opened 2023-09-06
add the auth logout command

gajicdev:
Closes https://gitlab.nine.ch/gcp/mono-infra/-/issues/674

### DIFF
--- a/auth/logout.go
+++ b/auth/logout.go
@@ -1,0 +1,101 @@
+package auth
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/gob"
+	"encoding/hex"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+	"strings"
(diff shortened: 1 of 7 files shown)


auth/logout.go
```diff
@@ -0,0 +1,101 @@
...
+	filename, err := computeFilename(key)
+	if err != nil {
+		return err
+	}
+
+	if err := os.Remove(path.Join(homedir.HomeDir(), api.DefaultTokenCachePath, filename)); err != nil {
+		return fmt.Errorf("error removing the local cache: %w", err)
+	}
+
+	format.PrintSuccessf("👋", "logged out from nctl")
```
pawelkuc, line 70:
```suggestion
	format.PrintSuccessf("👋", "logged out from %s", l.APIURL)
```

gajicdev: added

PR #59, opened 2023-09-06
Add whoami information on successful login

pawelkuc:
Display `whoami` information on successful login. It displays user name for an OIDC login, and Service Account name for a static API token login.

### DIFF
--- a/auth/login.go
+++ b/auth/login.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/ninech/nctl/api"
 	"github.com/ninech/nctl/api/util"
@@ -20,7 +21,6 @@ type LoginCmd struct {
 	Organization string `help:"The name of your organization to use when providing an API token. This parameter is only used when providing a API token. This parameter needs to be set if you use --api-token." env:"NCTL_ORGANIZATION"`
 	IssuerURL    string `help:"Issuer URL is the OIDC issuer URL of the API." default:"https://auth.nine.ch/auth/realms/pub"`
 	ClientID     string `help:"Client ID is the OIDC client ID of the API." default:"nineapis.ch-f178254"`
-	ExecPlugin   bool   `help:"Automatically run exec plugin after writing the kubeconfig." hidden:"" default:"true"`
 }
 
 func (l *LoginCmd) Run(ctx context.Context, command string, tk api.TokenGetter) error {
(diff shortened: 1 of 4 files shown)


auth/login.go
```diff
@@ -226,20 +226,36 @@ func login(ctx context.Context, newConfig *clientcmdapi.Config, kubeconfigPath,
...
+		if err != nil {
 			return err
 		}
+	}
 
+	userInfo, err := api.GetUserInfoFromToken(token)
+	if err != nil {
 		format.PrintSuccessf("🚀", "logged into cluster %s", newConfig.CurrentContext)
+
+		return nil
```
ctrox, line 255: Nit: Can you add a comment on why we don't return the error here? Are there any known cases where the user info can't be extracted from a valid token?

pawelkuc: yeah. It doesn't look good. I agree. However we have one real use case when there is no `UserInfo` on login available
```bash
nctl auth cluster <name>
# --exec-plugin=false by default
```

And even more cases in our unit tests:
1. auth/cluster_test.go
2. auth/login_test.go (TestLoginCmd, TestLoginCmdWithoutExistingKubeconfig)

assuming login() returns error on missing UserInfo, for the 2. case I was able to create the following workaround (added 'Organization` and `APIToken` to login command:
```golang
	// we run without the execPlugin, that would be something for an e2e test.
	// For this reason we also need valid Organization and APIToken fields to
	// pass GetUserInfoFromToken() in auth/login.go.
	cmd := &LoginCmd{
		ExecPlugin:   false,
		APIURL:       "https://" + apiHost,
		IssuerURL:    "https://auth.example.org",
		Organization: "test",
		APIToken:     FakeJWTToken,
	}
``` 
so this could work. 

But we still have `ClusterCmd` which accepts false for `ExecPlugin`, which means missing `UserInfo`, which means error on login.

Ignoring error on `GetUserInfoFromToken()` is the simplest solution that fixes everything. In the end, only information is missing - we do not use `UserInfo` for further calculations.

But maybe you have another idea how to solve this issue?

pawelkuc: as you suggested today, I simplified the login logic a bit. This solves all the problems described above.

auth/set_org.go
```diff
@@ -36,5 +35,7 @@ func (s *SetOrgCmd) Run(ctx context.Context, command string) error {
 		return err
 	}
 
-	return login(ctx, cfg, loadingRules.GetDefaultFilename(), s.Organization, runExecPlugin(s.ExecPlugin), project(s.Organization))
+	userInfo := &api.UserInfo{}
```
pawelkuc, line 38: A placeholder for the future. There is a bug in the current implementation of the `set-org` command (nctl auth set-org overwrites kubeconfig with OIDC login) and this block of code will likely be redesigned.

auth/login.go
ctrox: Looks good 👍

ctrox (approved):

PR #60, opened 2023-09-06
Logs for build since creation and for app since one hour

gajicdev:
Closes https://gitlab.nine.ch/gcp/mono-infra/-/issues/812

### DIFF
--- a/logs/build.go
+++ b/logs/build.go
@@ -2,7 +2,9 @@ package logs
 
 import (
 	"context"
+	"time"
 
+	apps "github.com/ninech/apis/apps/v1alpha1"
 	"github.com/ninech/nctl/api"
 )
 
@@ -13,6 +15,14 @@ type buildCmd struct {
 }
 
 func (cmd *buildCmd) Run(ctx context.Context, client *api.Client) error {
+	var build *apps.Build
+	if cmd.Name != "" {
+		if err := client.Get(ctx, api.NamespacedName(cmd.Name, client.Project), build); err != nil {
+			return err
+		}
+		cmd.Since = time.Since(build.CreationTimestamp.Time)
+	}
+
 	query := BuildQuery(cmd.Name, client.Project)
 	if len(cmd.ApplicationName) != 0 {
 		query = queryString(map[string]string{
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -20,8 +20,8 @@ type Cmd struct {
 
 type logsCmd struct {
 	Follow bool          `help:"Follow the logs by live tailing." short:"f"`
-	Lines  int           `help:"Amount of lines to output" default:"10" short:"l"`
-	Since  time.Duration `help:"Duration how long to look back for logs" short:"s" default:"10m"`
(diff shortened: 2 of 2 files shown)


logs/logs.go
```diff
@@ -19,10 +19,10 @@ type Cmd struct {
 }
 
 type logsCmd struct {
-	Follow bool          `help:"Follow the logs by live tailing." short:"f"`
-	Lines  int           `help:"Amount of lines to output" default:"10" short:"l"`
-	Since  time.Duration `help:"Duration how long to look back for logs" short:"s" default:"10m"`
-	Output string        `help:"Configures the log output format. ${enum}" short:"o" enum:"default,json" default:"default"`
+	Follow bool           `help:"Follow the logs by live tailing." short:"f"`
+	Lines  int            `help:"Amount of lines to output" default:"10" short:"l"`
+	Since  *time.Duration `help:"Duration how long to look back for logs" short:"s" default:"60m"`
```
ctrox, line 24: Even though this is a pointer, kong will still default to 60m. I think we have to use a different approach. We should use the `Since` value whenever the build name is not specified. If the build name is specified I don't think there's ever a reason to query logs before the build as there won't be anything there anyway.

logs/build.go
```diff
@@ -13,6 +15,11 @@ type buildCmd struct {
 }
 
 func (cmd *buildCmd) Run(ctx context.Context, client *api.Client) error {
+	var build apps.Build
+	if err := client.Get(ctx, api.NamespacedName(cmd.Name, client.Project), &build); err != nil {
+		return err
+	}
```
ctrox, line 21: This will break querying build logs by app name. So we need to first check if `cmd.Name` is not empty.

logs/build.go
```diff
@@ -21,6 +28,8 @@ func (cmd *buildCmd) Run(ctx context.Context, client *api.Client) error {
 		}, client.Project)
 	}
 
+	since := time.Since(build.CreationTimestamp.Time)
+	cmd.Since = &since
```
ctrox, line 32: Same with the build get, this only works if we actually have a build name.

logs/logs.go
gajicdev: changed it back to a normal value 👍

logs/build.go
gajicdev: added a check

logs/build.go
gajicdev: added a check here aswell

logs/build.go
```diff
@@ -21,6 +30,10 @@ func (cmd *buildCmd) Run(ctx context.Context, client *api.Client) error {
 		}, client.Project)
 	}
 
+	if build != nil {
+		cmd.Since = time.Since(build.CreationTimestamp.Time)
```
ctrox, line 34: Nit: can't this just be moved up into the `if cmd.Name != "" {` block after the `Get`? Then you don't need to nil check the `build`.

gajicdev: yup 🤦

PR #61, opened 2023-09-07
Update DNS verification mechanism

thirdeyenick:
As we recently introduced a TXT record domain verification, `nctl` now needs to display the needed TXT record for this. It will also link to a help article which explains the DNS setup for custom host names.

Here is an example:
```text
nctl create application go-example \
  --git-url=https://github.com/ninech/deploio-examples \
  --git-sub-path=go \
  --hosts=go-example.nine.ch,go-example-2.nine.ch

Creating a new application
 ✓ created application "go-example" 🏗
 ✓ waiting for build to start ⏳
 ✓ building application 📦
 ✓ releasing application 🚦
 ✓ release available ⛺

Your application "go-example" is now available at:
  https://go-example.7f0b629.deploio.app

You configured the following hosts:
  go-example-2.nine.ch
  go-example.nine.ch

Your DNS details are:
  TXT record:	deploio-site-verification=go-example-nick-8335ce6
  CNAME target:	go-example.7f0b629.deploio.app

To make your app available on your custom hosts, please use
the DNS details and visit https://docs.nine.ch/a/myshbw3EY1
for further instructions.
```


get/application.go
```diff
@@ -75,7 +80,7 @@ func printApplication(apps []apps.Application, get *Cmd, out io.Writer, header b
 
 func printCredentials(creds []appCredentials, get *Cmd, out io.Writer) error {
 	if get.Output == yamlOut {
-		return format.PrettyPrintObject(creds, format.PrintOpts{Out: out})
+		return format.PrettyPrintObjects(creds, format.PrintOpts{Out: out})
```
ctrox, line 83: Unrelated to this change but just got me thinking that we should probably detect if we're in a terminal within `PrettyPrintObjects` and then output the yaml object without any color codes so it's actually usable for scripting.

ctrox (approved):

get/application.go
thirdeyenick: You are actually right. This would make sense.

just FYI: This change was only done as I preferred the output in multiple yaml documents instead of showing a sequence directly. Not sure if this actually is the right way (or whats better for scripting). In the end I just want them to be consistent between showing basic auth credentials and the dns details.

thirdeyenick: I add the terminal handling part. All writes to a non-terminal are now made without colour.

PR #62, opened 2023-09-11
dont overwrite env vars when updating

gajicdev:
Closes https://gitlab.nine.ch/gcp/mono-infra/-/issues/800


api/util/apps.go
```diff
@@ -66,6 +67,37 @@ func EnvVarsFromMap(env map[string]string) apps.EnvVars {
...
+	}
+
+	if newEnvs != nil {
+		new := EnvVarsFromMap(*newEnvs)
+		for _, v := range new {
+			envMap[v.Name] = v
+		}
+	}
+
+	if toDelete != nil {
```
thirdeyenick, line 83: I'd just iterate over the slice and not make it a `*[]string`, but instead a normal `[]string`. Iterating over a nil slice is possible without a panic: https://go.dev/play/p/M1ntG_4FzON.

api/util/apps.go
```diff
@@ -66,6 +67,37 @@ func EnvVarsFromMap(env map[string]string) apps.EnvVars {
 	return vars
 }
 
+func UpdateEnvVars(oldEnvs []apps.EnvVar, newEnvs *map[string]string, toDelete *[]string) apps.EnvVars {
+	envMap := map[string]apps.EnvVar{}
+	for _, v := range oldEnvs {
+		envMap[v.Name] = v
+	}
+
+	if newEnvs != nil {
```
thirdeyenick, line 76: I'd also just use a `map[string]string` here and no pointer to it.  You can just iterate over it if needed.

api/util/apps.go
gajicdev: I used a pointer since its easier to pass the arguments from application.go in the update package:

`util.UpdateEnvVars(app.Spec.ForProvider.Config.Env, cmd.Env, cmd.DeleteEnv)`

if I derefence like `*cmd.Env`, I would need to nil check

gajicdev: switched away from pointers 👍

api/util/apps.go
gajicdev: here too

thirdeyenick (approved):
👍

PR #63, opened 2023-09-12
Add app build retries

pawelkuc:
In case a build fails, there's no idiomatic way to retry it. Sometimes the failure is temporary, such as an image upload that fails and a simple retry would fix it. There's a workaround that works well but is maybe not the most user-friendly: 
```bash
`nctl update app <app> --build-env=BUILD_TRIGGER=$(date)`
```
A very pragmatic solution is to simply hide this behind a retry-build flag. And this PR does just that.

### DIFF
--- a/update/application.go
+++ b/update/application.go
@@ -12,6 +12,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// BuildTrigger is used to request a retry-build for the application.
+const BuildTrigger = "BUILD_TRIGGER"
+
 // all fields need to be pointers so we can detect if they have been set by
 // the user.
 type applicationCmd struct {
@@ -27,6 +30,7 @@ type applicationCmd struct {
 	BuildEnv       *map[string]string `help:"Environment variables names which are passed to the app build process."`
 	DeleteBuildEnv *[]string          `help:"Build environment variables which are to be deleted."`
 	DeployJob      *deployJob         `embed:"" prefix:"deploy-job-"`
+	RetryBuild     *bool              `help:"Retries build for the application if set to true." placeholder:"false"`
 }
 
 type gitConfig struct {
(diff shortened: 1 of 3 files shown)


update/application.go
```diff
@@ -141,6 +145,16 @@ func (cmd *applicationCmd) applyUpdates(app *apps.Application) {
 	if cmd.Env != nil {
```
thirdeyenick, line 145: I know this has nothing to do with your changes, but this statement seems to be wrong? It should be `if cmd.BuildEnv != nil`, right?

update/application.go
```diff
@@ -141,6 +145,16 @@ func (cmd *applicationCmd) applyUpdates(app *apps.Application) {
...
 		buildEnv = *cmd.BuildEnv
 	}
+
+	if cmd.RetryBuild != nil {
+		if *cmd.RetryBuild {
+			if buildEnv == nil {
```
thirdeyenick, line 151: This somehow looks a bit too nested :-) I would do a `buildEnv := make(map[string]string)` on line 144 and then you can simplify this:

```go
if cmd.RetryBuild != nil && *cmd.RetryBuild {
  buildEnv[BuildTrigger] = time.Now().UTC().Format(time.RFC3339)
}
```

update/application_test.go
```diff
@@ -109,6 +109,7 @@ func TestApplication(t *testing.T) {
 					Command: pointer.String("exit 0"), Name: pointer.String("exit"),
 					Retries: pointer.Int32(1), Timeout: pointer.Duration(time.Minute * 5),
 				},
+				// RetryBuild should be not set by default
```
thirdeyenick, line 112: This comment should be closer to line 129 maybe?

update/application.go
pawelkuc: yeah. good catch! 👌

thirdeyenick (approved):
👍

PR #64, opened 2023-09-19
Fix set-org command

pawelkuc:
This PR fixes the `nctl auth set-org` command so that it no longer overrides the login method currently defined in .kube/config (static API token or OIDC) with OIDC.

It also adds a small feature to display a list of available organizations when an organization name is not specified (only for OIDC login method).

### DIFF
--- a/auth/config.go
+++ b/auth/config.go
@@ -141,9 +141,46 @@ func SetContextProject(kubeconfigPath string, contextName string, project string
 	}
 	context, exists := kubeconfig.Contexts[contextName]
 	if !exists {
-		return fmt.Errorf("could not find cluster %q in kubeconfig", contextName)
+		return fmt.Errorf("could not find context %q in kubeconfig", contextName)
 	}
 	context.Namespace = project
+	return clientcmd.WriteToFile(*kubeconfig, kubeconfigPath)
+}
+
+// SetContextOrganization sets the given organization in the given context of the kubeconfig
+func SetContextOrganization(kubeconfigPath string, contextName string, organization string) error {
+	kubeconfig, err := clientcmd.LoadFromFile(kubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("kubeconfig not found: %w", err)
+	}
+	context, exists := kubeconfig.Contexts[contextName]
+	if !exists {
+		return fmt.Errorf("could not find context %q in kubeconfig", contextName)
+	}
+	extension, exists := context.Extensions[util.NctlName]
+	if !exists {
+		return ErrConfigNotFound
+	}
+
+	cfg, err := parseConfig(extension)
+	if err != nil {
+		return err
+	}
+
+	if cfg.Organization == organization {
+		return nil
+	}
+
+	cfg.Organization = organization
+	cfgObject, err := cfg.ToObject()
+	if err != nil {
+		return err
+	}
+	context.Extensions[util.NctlName] = cfgObject
+
+	// change project to default for the the given organization:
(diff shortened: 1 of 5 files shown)


auth/config.go
```diff
@@ -147,6 +147,39 @@ func SetContextProject(kubeconfigPath string, contextName string, project string
...
 
+// SetContextOrganization sets the given organization in the given context of the kubeconfig
+func SetContextOrganization(kubeconfigPath string, contextName string, organization string) error {
+	kubeconfig, err := clientcmd.LoadFromFile(kubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("kubeconfig not found: %w", err)
+	}
+	context, exists := kubeconfig.Contexts[contextName]
+	if !exists {
+		return fmt.Errorf("could not find cluster %q in kubeconfig", contextName)
```
thirdeyenick, line 158:
```suggestion
		return fmt.Errorf("could not find context %q in kubeconfig", contextName)
```

auth/config.go
```diff
@@ -147,6 +147,39 @@ func SetContextProject(kubeconfigPath string, contextName string, project string
...
+	if !exists {
+		return ErrConfigNotFound
+	}
+
+	cfg, err := parseConfig(extension)
+	if err != nil {
+		return err
+	}
+
+	cfg.Organization = organization
```
thirdeyenick, line 170: Maybe we could exit here already if the organization which we want to set is already set? That way we wouldn't overwrite the `context.Namespace` later if the user already set this to be something else.

auth/whoami.go
```diff
@@ -38,13 +37,16 @@ func printUserInfo(userInfo *api.UserInfo, cfg *Config) {
 
 	fmt.Printf("Your current organization: %q\n", cfg.Organization)
 
-	if len(userInfo.Orgs) > 1 {
+	// TODO: temporarily changed. Do we need to redesign this block?
```
thirdeyenick, line 40: Fine with me to show this if there is only one organization available. We should just remove the comment.

auth/whoami.go
```diff
@@ -38,13 +37,16 @@ func printUserInfo(userInfo *api.UserInfo, cfg *Config) {
...
+	if len(userInfo.Orgs) > 0 {
 		printAvailableOrgsString(cfg.Organization, userInfo.Orgs)
 	}
 }
 
 func printAvailableOrgsString(currentorg string, orgs []string) {
-	fmt.Print("\nActive\tOrganization\n")
+	// TODO: temporarily changed. Do we need to redesign this block?
```
thirdeyenick, line 47: Also fine with. Just remove the comment.

main.go
```diff
@@ -110,11 +110,6 @@ func main() {
 		return
 	}
 
-	if strings.HasPrefix(kongCtx.Command(), format.SetOrgCommand) {
```
thirdeyenick, line 113: Why did we actually remove this?

auth/config.go
pawelkuc: yeah - very good idea! 👌

main.go
pawelkuc: this block of commands was design to:

> handle the login/oidc cmds separately as we should not try to get the API client if we're not logged in.

The previous version of `set-org` command performed the login operation for us. But we don't need this anymore. We just use a normal `Run()` with `client *api.Client` as argument. And we need `client` to read kubeconfig path or eventually (if org argument is missing) forward the request to `whoami` command which also needs `client`.

PR #68, opened 2023-11-06
feat: add deploy-job to config create/update

ctrox:
This makes it possible to configure deploy-jobs with configs.

mono-infra#845

### DIFF
--- a/api/util/apps.go
+++ b/api/util/apps.go
@@ -21,6 +21,7 @@ const (
 	// DNSSetupURL redirects to the proper deplo.io docs entry about
 	// how to setup custom hosts
 	DNSSetupURL = "https://docs.nine.ch/a/myshbw3EY1"
+	NoneText    = "<none>"
 )
 
 func UnverifiedAppHosts(app *apps.Application) []string {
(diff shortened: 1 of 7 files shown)


pawelkuc:
Just wondering if we need to show some `deploy-job` info (e.g. name of the job if there is a job configured) when we `Get deplo.io Project Configuration`. For now the get command for config gives us:
```
"NAME",
"SIZE",
"REPLICAS",
"PORT",
"ENVIRONMENT_VARIABLES",
"AGE",
```
which is almost everything we store in the `Config struct{}` minus `BasicAuth` and `DeployJob`.

pawelkuc (approved):

ctrox:
> Just wondering if we need to show some `deploy-job` info (e.g. name of the job if there is a job configured) when we `Get deplo.io Project Configuration`.

Was also thinking about that but a deploy-job has 4 different attributes, so adding 4 more columns would make the already wide output incredibly wide 😄

ctrox:
@pawelkuc I added a commit, it now looks like this:

```bash
$ nctl get config
NAME    SIZE     REPLICAS    PORT    ENVIRONMENT_VARIABLES    BASIC_AUTH    DEPLOY_JOB    AGE
nine    micro    0           8080    <none>                   false         <none>        75d
```

pawelkuc:
looks good! thank you 😄

PR #70, opened 2023-12-06
delete git auth secret when app creation fails & do not create git auth secret if it already exists

gajicdev:
Related to: 

https://gitlab.nine.ch/ninech/platform/mono-infra/-/issues/896
https://gitlab.nine.ch/ninech/platform/mono-infra/-/issues/897

In addition, this PR upgrades nctl to go 1.21

### DIFF
--- a/create/application.go
+++ b/create/application.go
@@ -120,6 +120,13 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
 	defer cancel()
 
 	if err := c.createResource(appWaitCtx); err != nil {
+		if auth.Enabled() {
+			secret := auth.Secret(newApp)
+			if gitErr := client.Delete(ctx, secret); err != nil {
+				return errors.Join(err, fmt.Errorf("unable to delete git auth secret, please contact support: %w", gitErr))
+			}
+		}
+
 		return err
 	}
 
--- a/create/application.go
+++ b/create/application.go
@@ -19,6 +19,7 @@ import (
 	"github.com/ninech/nctl/api/util"
 	"github.com/ninech/nctl/internal/format"
 	"github.com/ninech/nctl/logs"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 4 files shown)


create/application.go
```diff
@@ -102,10 +103,18 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
 			return fmt.Errorf("the credentials are given but they are empty: %w", err)
 		}
-		// for git auth we create a separate secret and then reference it in the app.
+
 		secret := auth.Secret(newApp)
-		if err := client.Create(ctx, secret); err != nil {
-			return fmt.Errorf("unable to create git auth secret: %w", err)
+		if err := client.Get(ctx, types.NamespacedName{Name: secret.Name, Namespace: secret.Namespace}, secret); err != nil {
+			// for git auth we create a separate secret and then reference it in the app.
+			if err := client.Create(ctx, secret); err != nil {
```
ctrox, line 110: I guess this should only happen when `if !errors.IsNotFound(err) {`?

ctrox: Or rather it might be easier to do `client.Create` and if that fails with `errors.IsAlreadyExists` we do the update.

create/application.go
```diff
@@ -102,10 +103,18 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
+				return fmt.Errorf("unable to create git auth secret: %w", err)
+			}
+		} else {
+			fmt.Println("updating git auth credentials")
+			if err := client.Update(ctx, secret); err != nil {
```
ctrox, line 115: It's a bit dangerous to just update the secret and discard whatever was in there. The user could create a secret on the API with the same name, unrelated to this app and then have his data overwritten. So we should check if the secret has the `ManagedByAnnotation: nctl` and only then update it.

ctrox (changes requested):

create/application.go
gajicdev: changed it to `errors.IsAlreadyExists` variant

create/application.go
gajicdev: added

create/application.go
```diff
@@ -102,10 +103,21 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
 		}
-		// for git auth we create a separate secret and then reference it in the app.
+
 		secret := auth.Secret(newApp)
+		// for git auth we create a separate secret and then reference it in the app.
 		if err := client.Create(ctx, secret); err != nil {
-			return fmt.Errorf("unable to create git auth secret: %w", err)
+			if k8sError.IsAlreadyExists(err) {
+				// only update the secret if it is managed by nctl in the first place
+				if _, exists := newApp.Annotations[util.ManagedByAnnotation]; exists {
```
ctrox, line 112: this only checks if the managed by annotation key is there but the value is also important :) it should also check if the value is `util.NctlName`.

```suggestion
				if v, exists := newApp.Annotations[util.ManagedByAnnotation]; exists && v == util.NctlName {
```

create/application.go
```diff
@@ -102,10 +103,21 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
+					fmt.Println("updating git auth credentials")
+					if err := client.Update(ctx, secret); err != nil {
```
ctrox, line 114: Not 100% sure on this but can you just call `Update` without getting the resource first? See `update/application.go` where we already do this. Also might be worth adding a test here.

create/application.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/ninech/nctl/api/util"
 	"github.com/ninech/nctl/internal/format"
 	"github.com/ninech/nctl/logs"
+	k8sError "k8s.io/apimachinery/pkg/api/errors"
```
ctrox, line 22: Nit: would name this `kerrors` for consistency (not yet in this repo but our mono-infra uses it like that)

ctrox (approved):

create/application.go
gajicdev: done

PR #72, opened 2023-12-22
fix: update dependencies

ctrox:
this mainly upgrades github.com/ninech/apis and loki because of some broken dependencies. The loki upgrade required some changes to the fake client since the interface changed slightly. The other test changes are due to a behaviour changes of the k8s fake client.


api/client.go
```diff
@@ -56,12 +55,8 @@ func New(ctx context.Context, apiClusterContext, project string, opts ...ClientO
 		return nil, err
 	}
 
-	mapper := apis.StaticRESTMapper(scheme)
```
ctrox, line 59: I have removed this since we now use the lazy mapper that comes with the new controller-runtime.

create/application_test.go
```diff
@@ -411,7 +411,7 @@ func TestApplicationWait(t *testing.T) {
 				app.Status.AtProvider.Hosts = []apps.VerificationStatus{{Name: "host.example.org"}}
 				app.Status.AtProvider.CNAMETarget = "some.target.example.org"
 				app.Status.AtProvider.BasicAuthSecret = &meta.LocalReference{Name: basicAuth.Name}
-				if err := apiClient.Status().Update(ctx, app); err != nil {
```
ctrox, line 414: Not sure what is going on with the fake client but updating the status subresource does not do anything but just calling `Update()` seems to update both the spec and the status. So the tests are now passing with just not using `Status().Update` anymore.

gajicdev: super weird 😅

ctrox:
@gajicdev FYI I added another commit to replace the deprecated `k8s.io/utils/pointer` package as the linter complained.

PR #73, opened 2024-01-19
feat: output build logs during app create

ctrox:
this adds a log view during the initial build so you can see how the build is progressing without having to open a second terminal.

Here's how it looks:

[asciicast](https://asciinema.org/a/AkjBgRRsFTvLntiC5SQ3CVTWm)


internal/logbox/logbox.go
```diff
@@ -0,0 +1,143 @@
...
+		lb.spinner, cmd = lb.spinner.Update(msg)
+		return lb, cmd
+	default:
+		return lb, nil
+	}
+}
+
+func (lb LogBox) View() string {
+	var s string
+
```
thirdeyenick, line 90: Do we need all these empty lines in this function?

internal/logbox/logbox.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (f *Output) WithWriter(w io.Writer) output.LogOutput {
+	return f
+}
+
+// delaySend delays the sending of the message by the specified duration.
+func (f *Output) delaySend(d time.Duration, msg Msg) {
+	next := make(chan bool)
+	go func() {
```
thirdeyenick, line 134: Is there a specific reason why we wait in a parallel goroutine here instead of just directly in the func itself?

internal/logbox/logbox.go
ctrox: I think it makes things more readable :) but I removed the one after the var declaration

internal/logbox/logbox.go
ctrox: Yeah, that might be a case of over-engineering it :smile: I was at first worried that blocking might cause issues in the tail client but it should not mind such a small delay. I removed it for now, we can always reintroduce it if we experience any problems.

thirdeyenick (approved):

PR #75, opened 2024-02-01
use git information service on app creation and updates

thirdeyenick:
This implements the usage of the git information service into `nctl`. It allows to check if the specified git repository and revision can be accessed/found.

### DIFF
--- a/api/util/apps.go
+++ b/api/util/apps.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"encoding/pem"
 	"fmt"
 	"sort"
 	"strings"
@@ -128,6 +129,14 @@ type GitAuth struct {
 	SSHPrivateKey *string
 }
 
+func (git GitAuth) HasPrivateKey() bool {
+	return git.SSHPrivateKey != nil
+}
+
+func (git GitAuth) HasBasicAuth() bool {
+	return git.Username != nil && git.Password != nil
+}
+
 func (git GitAuth) Secret(app *apps.Application) *corev1.Secret {
 	data := map[string][]byte{}
 
@@ -170,16 +179,12 @@ func (git GitAuth) UpdateSecret(secret *corev1.Secret) {
 	secret.Annotations[ManagedByAnnotation] = NctlName
 }
 
+// Enabled returns true if any kind of credentials are set in the GitAuth
 func (git GitAuth) Enabled() bool {
-	if git.Username != nil ||
-		git.Password != nil ||
-		git.SSHPrivateKey != nil {
-		return true
-	}
-
-	return false
+	return git.HasBasicAuth() || git.HasPrivateKey()
 }
 
+// Valid validates the credentials in the GitAuth
 func (git GitAuth) Valid() error {
 	if git.SSHPrivateKey != nil {
 		if *git.SSHPrivateKey == "" {
@@ -228,4 +233,19 @@ func GatherDNSDetails(items []apps.Application) []DNSDetail {
 		result[i] = data
 	}
 	return result
+}
+
+// ValidatePEM validates if the passed content is in valid PEM format, errors
+// out if the content is empty
+func ValidatePEM(content string) (*string, error) {
+	if content == "" {
+		return nil, fmt.Errorf("the SSH private key cannot be empty")
+	}
+
+	content = strings.TrimSpace(content)
+	b, rest := pem.Decode([]byte(content))
+	if b == nil || len(rest) > 0 {
+		return nil, fmt.Errorf("no valid PEM formatted data found")
+	}
+	return &content, nil
 }
--- a/create/application.go
+++ b/create/application.go
@@ -2,7 +2,6 @@ package create
 
 import (
 	"context"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"os"
@@ -19,6 +18,7 @@ import (
 	"github.com/ninech/nctl/api"
 	"github.com/ninech/nctl/api/log"
 	"github.com/ninech/nctl/api/util"
+	"github.com/ninech/nctl/api/validation"
 	"github.com/ninech/nctl/internal/format"
 	"github.com/ninech/nctl/internal/logbox"
 	"github.com/ninech/nctl/logs"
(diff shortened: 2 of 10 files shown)


create/application.go
```diff
@@ -32,18 +31,21 @@ import (
...
+	Port                     *int32            `help:"Port the app is listening on (defaults to ${app_default_port})." placeholder:"${app_default_port}"`
+	Replicas                 *int32            `help:"Amount of replicas of the running app (defaults to ${app_default_replicas})." placeholder:"${app_default_replicas}"`
+	Hosts                    []string          `help:"Host names where the app can be accessed. If empty, the app will just be accessible on a generated host name on the deploio.app domain."`
+	BasicAuth                *bool             `help:"Enable/Disable basic authentication for the app (defaults to ${app_default_basic_auth})." placeholder:"${app_default_basic_auth}"`
+	Env                      map[string]string `help:"Environment variables which are passed to the app at runtime."`
+	BuildEnv                 map[string]string `help:"Environment variables which are passed to the app build process."`
+	DeployJob                deployJob         `embed:"" prefix:"deploy-job-"`
+	GitInformationServiceURL string            `help:"URL of the git information service." default:"https://git-info.deplo.io" env:"GIT_INFORMATION_SERVICE_URL"`
```
ctrox, line 46: I would hide this from the help with hidden:"" as I don't think any users need to override this?

create/application.go
```diff
@@ -102,6 +104,20 @@ func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
...
+		gitInfoClient, err := util.NewGitInformationClient(app.GitInformationServiceURL, client.Token)
+		if err != nil {
+			return err
+		}
+		gitInfoClient.SetLogRetryFunc(util.RetryLogFunc(app.Debug))
+		fmt.Println("Testing repository access")
+		if err := util.TestRepositoryAccess(ctx, gitInfoClient, &newApp.Spec.ForProvider.Git, auth); err != nil {
+			return err
+		}
+		fmt.Println("Repository access successfully tested")
```
ctrox, line 118: I gave this a spin locally and it would IMO look quite a bit nicer if this would make use of our spinner system. It's quite easy to use:

```golang
msg := message{text: "testing repository access", icon: "🔐"}
spinner, err := format.NewSpinner(msg.progress(), msg.progress())
if err != nil {
	return err
}

spinner.Start()
if err := doWork(); err != nil {
	spinner.StopFail()
	return err
}
spinner.Stop()
```

feel free to choose another icon, I just put something that sort of fits.

update/application.go
```diff
@@ -19,28 +20,46 @@ const BuildTrigger = "BUILD_TRIGGER"
...
+	Hosts                    *[]string          `help:"Host names where the application can be accessed. If empty, the application will just be accessible on a generated host name on the deploio.app domain."`
+	BasicAuth                *bool              `help:"Enable/Disable basic authentication for the application."`
+	Env                      *map[string]string `help:"Environment variables which are passed to the app at runtime."`
+	DeleteEnv                *[]string          `help:"Runtime environment variables names which are to be deleted."`
+	BuildEnv                 *map[string]string `help:"Environment variables names which are passed to the app build process."`
+	DeleteBuildEnv           *[]string          `help:"Build environment variables which are to be deleted."`
+	DeployJob                *deployJob         `embed:"" prefix:"deploy-job-"`
+	RetryBuild               *bool              `help:"Retries build for the application if set to true." placeholder:"false"`
+	GitInformationServiceURL string             `help:"URL of the git information service." default:"https://git-info.deplo.io" env:"GIT_INFORMATION_SERVICE_URL"`
```
ctrox, line 36: see other comment about this flag

api/util/apps.go
```diff
@@ -229,3 +241,51 @@ func GatherDNSDetails(items []apps.Application) []DNSDetail {
...
+		fmt.Fprintf(os.Stderr, "warning: %s\n", strings.Join(repoInfo.Warnings, "."))
+	}
+	if repoInfo.Error != "" {
+		return errors.New(repoInfo.Error)
+	}
+	if !(containsBranch(git.Revision, repoInfo) ||
+		containsTag(git.Revision, repoInfo)) {
+		return fmt.Errorf(
+			"can not find specified git revision (%q) in repository",
+			git.Revision,
```
ctrox, line 282: Would be kind of nice if this error would display the revisions it found in the repo. But I guess it could get quite long if there are lots of branches/tags :smile:

api/util/git_information_client.go
```diff
@@ -0,0 +1,206 @@
+package util
```
ctrox, line 1: Nit: wouldn't this fit better into a separate package like the log client which is in `api/log`?

internal/test/apps.go
```diff
@@ -0,0 +1,131 @@
...
+		Bytes: marshalPrivateED25519Key(privateKey, publicKey),
+	}
+
+	// Private key in PEM format
+	return strings.TrimSpace(string(pem.EncodeToMemory(pemKey))), nil
+}
+
+// Writes ed25519 private keys into the new OpenSSH private key format.
+// (Function taken and slightly modified from: https://github.com/mikesmitty/edkey/blob/master/edkey.go)
+func marshalPrivateED25519Key(privateKey ed25519.PrivateKey, publicKey ed25519.PublicKey) []byte {
```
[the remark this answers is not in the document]
thirdeyenick: Great stuff. This saves quite some lines :-)

create/application.go
thirdeyenick: I added the spinner. Looks better now indeed.

api/util/apps.go
thirdeyenick: Yeah was thinking about this too, but then in the end the important message is that the revision you specified is not in. It would be nice if something like "did you mean <suggested revision>" would be shown, but this is a bit out of scope I'd say.

create/application.go
thirdeyenick: Still need to do more DRY there...

thirdeyenick: Ok, as we use exactly the same for `create` and `update`, I moved the logic into a different package.

api/util/git_information_client.go
thirdeyenick: I moved quite some stuff into `api/validation` now.

api/util/apps.go
ctrox: yeah, fuzzy matching and then `did you mean x` would probably the best user experience. Agreed that it is out of scope for now.

ctrox (approved):

PR #77, opened 2024-02-06
fix: increase project deletion timeout

ctrox:
deleting a project can take a bit longer if it has a lot of resources in it that need to be deleted first. This simply increases the default timeout to 5 minutes.


thirdeyenick (approved):
👍

PR #78, opened 2024-02-20
fix: validate deploy job name

pawelkuc:
If the [deploy job](https://docs.nine.ch/docs/deplo-io/configuration/deploio-deploy-jobs) name is not set, an internal error will occur. This PR adds validation logic so that we can ensure that the deploy job name is provided if the deploy job command is specified.


api/validation/config.go
```diff
@@ -0,0 +1,22 @@
...
+	apps "github.com/ninech/apis/apps/v1alpha1"
+)
+
+// ConfigValidator validates a Config
+type ConfigValidator struct {
+	Config apps.Config
+}
+
+// Validate validates the config
+func (c ConfigValidator) Validate() error {
```
ctrox, line 15: Should we also copy over the other validations we have on the backend for the config (e.g env vars)?

ctrox (approved):

api/validation/config.go
ctrox: As discussed, it's fine if we merge this as is and add more validation later. We anyways have validation on the API side so it will just fall back to that.

PR #79, opened 2024-02-26
fix: do not use pointers for env maps

ctrox:
using pointers breaks kong's arg parsing somewhat as it fails to merge multiple values of the same arg. Removing the pointer fixes that while still making it possible to detect if the arg has been supplied or not as a map has nil as the zero value.

### DIFF
--- a/update/application.go
+++ b/update/application.go
@@ -21,22 +21,22 @@ const BuildTrigger = "BUILD_TRIGGER"
 // all fields need to be pointers so we can detect if they have been set by
 // the user.
 type applicationCmd struct {
-	Name                     *string            `arg:"" help:"Name of the application."`
-	Git                      *gitConfig         `embed:"" prefix:"git-"`
-	Size                     *string            `help:"Size of the app."`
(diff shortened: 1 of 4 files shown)


update/application.go
```diff
@@ -180,7 +180,7 @@ func (cmd *applicationCmd) applyUpdates(app *apps.Application) {
 
 	var env map[string]string
 	if cmd.Env != nil {
-		env = *cmd.Env
+		env = cmd.Env
```
thirdeyenick, line 183: Can't we then get rid of the `env` variable and use `cmd.Env` directly? The `env` variable will also be `nil` if `cmd.Env` is `nil`.

thirdeyenick: But we do this already. If `cmd.Env` is `nil` (which it is if it is not set, right?), then the `env` var is also `nil` (as we don't use `make`).

ctrox: Yeah you are right, we don't make the make there, we just do it for the `buildEnv`.

thirdeyenick (approved):
👍

PR #81, opened 2024-02-29
Improve revision checking

thirdeyenick:
As the git information service can now check for the existence of specific git revisions, we can make use of that feature in nctl.


api/validation/apps.go
```diff
@@ -62,8 +65,8 @@ func testRepositoryAccess(ctx context.Context, client *GitInformationClient, git
 	if repoInfo.Error != "" {
 		return errors.New(repoInfo.Error)
 	}
-	if !(containsBranch(git.Revision, repoInfo) ||
-		containsTag(git.Revision, repoInfo)) {
+	if repoInfo.RepositoryInfo.RevisionResponse != nil &&
+		!repoInfo.RepositoryInfo.RevisionResponse.Found {
 		return fmt.Errorf(
 			"can not find specified git revision (%q) in repository",
```
ctrox, line 71: do we still get back the list of branches if our requested branch does not exist? Just thinking it would be kind of cool if nctl said:

> could not find branch `main` in repository, did you mean `master`?

With some fuzzy matching to make suggestions. Maybe out of scope for this PR but we could make a ticket :)

ctrox (approved):

api/validation/apps.go
thirdeyenick: Yes, we still return the list of branches and tags in the repo even if the requested revision could not be found. As the git information service does not respond with an error message, but instead with a simple `Found: false` field, that fuzzy matching would need to be implemented on the client side I'd say. It is a nice idea, maybe for a freedom friday :-)

PR #112, opened 2024-06-14
refactor: move common flags to common struct

ctrox:
Certain flags are repeated within every resource. This commit refactors them to a common struct within each top-level command. This helps with consistency and will also help with defining autocompletion more centrally.


thirdeyenick (approved):
👍

PR #113, opened 2024-06-17
feat: add dynamic completion

ctrox:
This adds initial support for dynamic completion. For example `nctl update application <tab><tab>` will return a list of applications in the current project. There are some edge-cases where the arg/flag name does not match the resource name. To make these cases work, they can be added to a map in the predictor.

### DIFF
--- a/predictor/predictor.go
+++ b/predictor/predictor.go
@@ -1,0 +1,99 @@
+package predictor
+
+import (
+	"context"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/gobuffalo/flect"
+	"github.com/ninech/apis/management/v1alpha1"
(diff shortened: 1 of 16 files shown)


predictor/predictor.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return resources
+}
+
+func (r *Resource) findKind(arg string) schema.GroupVersionKind {
+	if v, ok := argResourceMap[arg]; ok {
+		arg = v
+	}
+
+	// fmt.Println(flect.Pluralize(arg))
```
thirdeyenick, line 84: guess that is a leftover...

thirdeyenick (approved):
👍

PR #122, opened 2024-07-05
set the docker API version to 1.42

Nemat97:
ninech/platform/mono-infra#1092

### DIFF
--- a/get/build.go
+++ b/get/build.go
@@ -22,6 +22,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/duration"
 )
 
+const (
+	dockerAPIVersion = "1.42"
+)
+
 type buildCmd struct {
 	resourceCmd
 	ApplicationName string `short:"a" help:"Name of the Application to get builds for. If omitted all in the project will be listed."`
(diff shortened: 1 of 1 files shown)


get/build.go
```diff
@@ -84,7 +88,7 @@ func printBuild(builds []apps.Build, get *Cmd, out io.Writer, header bool) error
 }
 
 func pullImage(ctx context.Context, apiClient *api.Client, build *apps.Build) error {
-	cli, err := client.NewClientWithOpts(client.FromEnv)
+	cli, err := client.NewClientWithOpts(client.WithVersion(dockerAPIVersion))
```
thirdeyenick, line 91: something which just came to my mind:
As we are leaving the `client.FromEnv` out now, we remove the possibility to overwrite the version or any other settings via environment variables.  Maybe we can use both options?

```go
cli, err := client.NewClientWithOpts(client.WithVersion(dockerAPIVersion), client.FromEnv)
```

This might set a default for the version to use which can then still be overwritten by env variables. Could you try that?

thirdeyenick (changes requested):

get/build.go
```diff
@@ -84,7 +88,12 @@ func printBuild(builds []apps.Build, get *Cmd, out io.Writer, header bool) error
 }
 
 func pullImage(ctx context.Context, apiClient *api.Client, build *apps.Build) error {
-	cli, err := client.NewClientWithOpts(client.FromEnv)
+	value, exists := os.LookupEnv("DOCKER_API_VERSION")
```
thirdeyenick, line 91: I think you actually don't need that. `client.WithVersion(dockerAPIVersion)` will set the hardcoded docker API version to 1.42. If the user specified an environment variable `DOCKER_API_VERSION`, then `client.FromEnv` should pick this up and overwrite our default version.

thirdeyenick (approved):

PR #124, opened 2024-07-10
fix: do not update/sort env vars when nothing changes

ctrox:
When updating an app with an unrelated change, we should not sort the existing env vars as that can trigger an unexpected build. We simply return the old env vars if the new envs and to delete is empty.


thirdeyenick (approved):
👍

PR #129, opened 2024-07-18
feat: add dockerfile flags

ctrox:
This adds all relevant flags for the app create/update command to enable and configure dockerfile build support.


update/application.go
```diff
@@ -223,6 +229,14 @@ func (cmd *applicationCmd) applyUpdates(app *apps.Application) {
 		buildDelEnv = *cmd.DeleteBuildEnv
 	}
 	app.Spec.ForProvider.BuildEnv = util.UpdateEnvVars(app.Spec.ForProvider.BuildEnv, buildEnv, buildDelEnv)
+
+	if cmd.DockerfileBuild.Path != nil {
```
thirdeyenick, line 233: Not sure, but do we want to output something like `Dockerfile builds are not enabled for this application, so setting the Dockerfile path will not have any affect` if Dockerfile builds are not enabled in the app? It is not necessary, but maybe a hint for the user?

ctrox: like just print a warning? We can't really error out in applyUpdates.

ctrox: I added a little warning:

```bash
$ go run . update app secure-corsair --dockerfile-build-context=bla
Warning: updating build context has no effect as dockefile builds are not enabled on this app
 ✓ updated Application "secure-corsair" ⬆️
```

thirdeyenick: Awesome! Looks good.

thirdeyenick (approved):

PR #131, opened 2024-07-23
allow to list only certain resource types

thirdeyenick:
This allows to only show certain resource types when listing all existing resources. It helps in finding resources without the need for a specific nctl implementation.

### DIFF
--- a/get/all.go
+++ b/get/all.go
@@ -22,7 +22,8 @@ import (
 type allCmd struct {
 	out                  io.Writer
 	stdErr               io.Writer
-	IncludeNineResources bool `help:"show resources which are owned by Nine" default:"false"`
+	Kinds                []string `help:"specify the kind of resources which should be listed"`
+	IncludeNineResources bool     `help:"show resources which are owned by Nine" default:"false"`
 }
 
 func (cmd *allCmd) Run(ctx context.Context, client *api.Client, get *Cmd) error {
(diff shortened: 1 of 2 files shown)


get/all.go
```diff
@@ -134,6 +145,26 @@ func printItems(items []*unstructured.Unstructured, get Cmd, out io.Writer, head
...
+		return lists, nil
+	}
+OUTER:
+	for _, kind := range kinds {
+		for _, list := range lists {
+			if !strings.EqualFold(kind+"list", list.GroupKind().Kind) {
+				continue
+			}
+			result = append(result, list)
+			continue OUTER
```
ctrox, line 161: isn't it simpler to just `break` here?

thirdeyenick: yeah, you are right 🤦

thirdeyenick: It actually doesn't work as the error gets thrown when we found a list type for a requested kind now. So we can either leave it with the `label` or we introduce a `bool` var which allows to find out if something was found. Wdyt?

get/all.go
```diff
@@ -70,11 +71,21 @@ func projectNames(projects []management.Project) []string {
 	return result
 }
 
-func getProjectContent(ctx context.Context, client *api.Client, projNames []string, includeNineOwned bool) ([]*unstructured.Unstructured, []string, error) {
+func getProjectContent(
```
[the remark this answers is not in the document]
thirdeyenick: Yeah good idea. Fixed it.

get/all.go
ctrox: Ah yeah makes sense. Both is fine with me honestly.

ctrox (approved):

get/all.go
thirdeyenick: I'll leave it then as it is. 👍

PR #133, opened 2024-07-24
fix: improve logbox interrupt handling

ctrox:
This fixes a number of issues with the cancel/error handling while the
logbox (tea program) is running. Previously input was sent to the tea
program, which meant we had to handle ctrl+c/ctrl+d explicitly. Now the
tea program is stopped using the context and also in afterWait since
it's important to call `p.Wait()` so that the console is restored
properly before nctl exits.

### DIFF
--- a/create/application.go
+++ b/create/application.go
@@ -12,7 +12,6 @@ import (
 	"github.com/alecthomas/kong"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/grafana/loki/pkg/logproto"
-	"github.com/mattn/go-isatty"
 	apps "github.com/ninech/apis/apps/v1alpha1"
 	meta "github.com/ninech/apis/meta/v1alpha1"
 	"github.com/ninech/nctl/api"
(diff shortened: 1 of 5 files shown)


create/application.go
```diff
@@ -360,17 +360,14 @@ func waitForBuildStart(app *apps.Application) waitStage {
...
+func waitForBuildFinish(ctx context.Context, app *apps.Application, logClient *log.Client) waitStage {
 	msg := message{icon: "📦", text: "building application"}
-	interrupt := make(chan bool, 1)
-	lb := logbox.New(15, msg.progress(), interrupt)
-	opts := []tea.ProgramOption{tea.WithoutSignalHandler()}
-
+	lb := logbox.New(15, msg.progress())
+	opts := []tea.ProgramOption{tea.WithoutSignalHandler(), tea.WithInput(nil), tea.WithContext(ctx)}
 	// disable input if we are not in a terminal
 	if !isatty.IsTerminal(os.Stdout.Fd()) {
```
thirdeyenick, line 368: I assume this block can be removed now.

thirdeyenick (approved):

PR #142, opened 2024-08-15
fix the update command with the language

gajicdev:
if you do not set the language explicitly, it will always be nil currenlty


ctrox (approved):
not sure what happened with the github actions (maybe give it a retry first) but the change itself looks good.

gajicdev:
@ctrox yeah not sure, locally the tests pass 🤷

ctrox:
@gajicdev I can see that the action `actions/setup-go@v5` downloads go 1.23 which just came out. We should pin the version to 1.22 and upgrade it later.

PR #147, opened 2024-08-28
fix: use default API cluster in autocomplete predictor

ctrox:
Instead of trying to call parse to default to the kong value, we now put the defualt API cluster into a const and use it in the default and when constructing the completion client. Additionally we check if the env var has been set to override the cluster.


thirdeyenick (approved):
👍

Issue #104, opened 2024-05-28
Following Ubuntu install instructions gives `Permission denied`


renepupil:
Obviously `echo "deb [trusted=yes] https://repo.nine.ch/deb/ /" > /etc/apt/sources.list.d/repo.nine.ch.list` does not work because of missing privileges to write to `sources.list.d` BUT `sudo echo "deb [trusted=yes] https://repo.nine.ch/deb/ /" > /etc/apt/sources.list.d/repo.nine.ch.list` also does NOT work because the superuser privilege ONLY applies to echo before the redirection.

Took me 5 minutes to check, will take new Devs probably a lot more time.

ALSO why do you recommend using an instable CLI?

```
$ apt list | head -n 10

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

This warning isn't there as a joke...

Please write the install guide in a way someone ACTUALLY had to execute the commands in a stable manner:

```
sudo sh -c 'echo "deb [trusted=yes] https://repo.nine.ch/deb/ /" > /etc/apt/sources.list.d/repo.nine.ch.list'
sudo apt-get update
sudo apt-get install nctl
```

**Can I create a PR for this?**

ctrox:
The apt

ctrox:
Sure, we're always welcoming PRs to improve code or docs.

I would suggest to use `| sudo tee` instead as it's a bit simpler to read than spawning a shell as root:

```bash
echo "deb [trusted=yes] https://repo.nine.ch/deb/ /" | sudo tee /etc/apt/sources.list.d/repo.nine.ch.list
```

While the apt interface is not stable, it's not like we're using it in a script. But yeah we can change it to `apt-get` to be safe.

renepupil:
@ctrox finally had time for the PR.

PR #149, opened 2024-08-28
fix: debian/ubuntu install instructions 2

renepupil:
#104 Add `sudo` to `apt-get` commands in `README.md`


ctrox (approved):
I missed those, thanks!

PR #157, opened 2024-09-05
adjust nctl output when searching for apps

Nemat97:
ninech/platform/mono-infra#1147

### DIFF
--- a/get/get.go
+++ b/get/get.go
@@ -224,7 +224,11 @@ func (cmd *Cmd) writeTabRow(w io.Writer, project string, row ...string) {
 	}
 }
 
-func printEmptyMessage(out io.Writer, kind, project string) {
+func (cmd *Cmd) printEmptyMessage(out io.Writer, kind, project string) {
+	if cmd.AllProjects {
+		fmt.Fprintf(defaultOut(out), "no %s found in any project\n", flect.Pluralize(kind))
+		return
+	}
 	if project == "" {
 		fmt.Fprintf(defaultOut(out), "no %s found\n", flect.Pluralize(kind))
 		return
--- a/get/get.go
+++ b/get/get.go
@@ -2,11 +2,13 @@ package get
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"reflect"
 	"slices"
+	"strings"
 
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/gobuffalo/flect"
@@ -35,6 +37,7 @@ type Cmd struct {
 	All                 allCmd                `cmd:"" name:"all" help:"Get project content"`
 	CloudVirtualMachine cloudVMCmd            `cmd:"" group:"infrastructure.nine.ch" name:"cloudvirtualmachine" aliases:"cloudvm" help:"Get a CloudVM."`
 	opts                []runtimeclient.ListOption
+	searchForName       string
 }
 
 type resourceCmd struct {
@@ -55,9 +58,11 @@ type listOpt func(cmd *Cmd)
 func matchName(name string) listOpt {
 	return func(cmd *Cmd) {
 		if len(name) == 0 {
+			cmd.searchForName = ""
 			return
 		}
 		cmd.opts = append(cmd.opts, runtimeclient.MatchingFields{"metadata.name": name})
+		cmd.searchForName = name
 	}
 }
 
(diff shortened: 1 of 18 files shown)


get/get.go
```diff
@@ -110,7 +168,70 @@ func (cmd *Cmd) list(ctx context.Context, client *api.Client, list runtimeclient
...
+	// we found the named object at least in one different project,
+	// so we return a hint to the user to search in these projects
+	var identifiedProjects []string
+	for i := 0; i < items.Len(); i++ {
+		// the "Items" field of a list type most often contains
+		// a slice of types, but not a slice of
+		// pointers to the type (e.g. "[]corev1.Pod" and not
+		// "[]*corev1.Pod"). We still want to check which
+		// variant (pointer or no pointer) implements the
+		// runtimeclient.Object interface.
```
ctrox, line 191: is this really needed? At least angryjet (crossplane gen) will already error out if items is a slice of pointers. I did not test any further but controller-runtime might also not like it.

get/get.go
```diff
@@ -110,7 +168,70 @@ func (cmd *Cmd) list(ctx context.Context, client *api.Client, list runtimeclient
...
+	if len(identifiedProjects) > 0 {
+		h := &hint{
+			prefix: "NOTE",
+			msg: fmt.Sprintf(
+				"we found the resource in the following project(s): %s\n"+
+					"Maybe you want to use the '--project' "+
+					"option to specify one of these projects?\n",
+				strings.Join(identifiedProjects, " ,"),
+			),
+		}
```
ctrox, line 211: Can't we attach the hint to a custom error type (or just an error)? IMO the current get behavior is already wrong:

```
$ nctl get app doesnotexist
no Applications found in project nine
```

The exit code is 0 but should be 1 and it should not say "no applications found", it should say "application x not found in project y". And then with the new hint system, it could just be in the error string after that text.

thirdeyenick: Sure, we can implement it like this. I was a bit hesitant to change already existing behavior, but I agree with your statements. `hint` is already a custom error, but would not be needed if we handle a 'your resource could not be found' error with exit code 1.

get/get.go
thirdeyenick: Yeah, I actually did not find any reference for this, where it says "`Items` always has to be a slice of non-pointer types". That is why we try both ways here. We can also drop the 'pointer' case and ignore an item if it is a pointer type.

get/get.go
ctrox: I'm just thinking that if you need to explicitly handle the hint, we'll forget about it when we add new types. With the error you are forced to handle it.

thirdeyenick: I adapted the code and implemented it as described above. Some examples:

```
$> ./nctl get app
PROJECT    NAME       HOSTS                               UNVERIFIED_HOSTS
nick       show-ip    show-ip.7b1b629.test.deploio.app    none

 $> ./nctl get app -A
PROJECT     NAME           HOSTS                                   UNVERIFIED_HOSTS
nick        show-ip        show-ip.7b1b629.test.deploio.app        none
nick-dev    dev-show-ip    dev-show-ip.d9d915b.test.deploio.app    none

 $> ./nctl get app show-ip
PROJECT    NAME       HOSTS                               UNVERIFIED_HOSTS
nick       show-ip    show-ip.7b1b629.test.deploio.app    none

 $> ./nctl get app dev-show-ip
nctl: error: resource "dev-show-ip" was not found in project nick, but it was found in project(s): nick-dev. Maybe you want to use the '--project' flag to specify one of these projects?

$> ./nctl get app dev-show-ip -A
PROJECT     NAME           HOSTS                                   UNVERIFIED_HOSTS
nick-dev    dev-show-ip    dev-show-ip.d9d915b.test.deploio.app    none

 $> ./nctl get mysql
no MySQLs found in project nick

$> ./nctl get mysql -A
no MySQLs found in any project
```

get/get.go
thirdeyenick: I am only checking for a non-pointer type now and skip the object if it is a pointer type.

ctrox (commented):
Playing around with this locally, the "found in other project" suggestions work great. But I'm not a 100% sure about putting the `PROJECT` column first. I know this is also how kubectl does it but I'm just not totally sure that at a glance the first thing you want to see is the project instead of the resouce name:

```bash
PROJECT    NAME                  HOSTS                              UNVERIFIED_HOSTS
foo        docs-page             docs-page.edcd8e7.deploio.app      none
foo        go-example            go-example.edcd8e7.deploio.app     none
# vs
NAME          PROJECT            HOSTS                              UNVERIFIED_HOSTS
docs-page     foo                docs-page.edcd8e7.deploio.app      none
go-example    foo                go-example.edcd8e7.deploio.app     none
```

thirdeyenick:
> Playing around with this locally, the "found in other project" suggestions work great. But I'm not a 100% sure about putting the `PROJECT` column first. I know this is also how kubectl does it but I'm just not totally sure that at a glance the first thing you want to see is the project instead of the resouce name:
> 
> ```shell
> PROJECT    NAME                  HOSTS                              UNVERIFIED_HOSTS
> foo        docs-page             docs-page.edcd8e7.deploio.app      none
> foo        go-example            go-example.edcd8e7.deploio.app     none
> # vs
> NAME          PROJECT            HOSTS                              UNVERIFIED_HOSTS
> docs-page     foo                docs-page.edcd8e7.deploio.app      none
> go-example    foo                go-example.edcd8e7.deploio.app     none
> ```

As the display function also is kind of generic, the best option we have is to either add the project to the most left or most right position. We could implement some logic which shows the `PROJECT` column always to the right of a possibly given `NAME` column, but I am not sure if it is worth the effort (and magic) :-)

ctrox:
> > Playing around with this locally, the "found in other project" suggestions work great. But I'm not a 100% sure about putting the `PROJECT` column first. I know this is also how kubectl does it but I'm just not totally sure that at a glance the first thing you want to see is the project instead of the resouce name:
> > ```shell
> > PROJECT    NAME                  HOSTS                              UNVERIFIED_HOSTS
> > foo        docs-page             docs-page.edcd8e7.deploio.app      none
> > foo        go-example            go-example.edcd8e7.deploio.app     none
> > # vs
> > NAME          PROJECT            HOSTS                              UNVERIFIED_HOSTS
> > docs-page     foo                docs-page.edcd8e7.deploio.app      none
> > go-example    foo                go-example.edcd8e7.deploio.app     none
> > ```
> 
> As the display function also is kind of generic, the best option we have is to either add the project to the most left or most right position. We could implement some logic which shows the `PROJECT` column always to the right of a possibly given `NAME` column, but I am not sure if it is worth the effort (and magic) :-)

Ah yeah, unfortunately the `NAME` column is always added manually in the specific get implementations. We should have just passed the whole `client.Object` to `get.writeTabRow`, then we would be way more flexible as we always have the name and namespace fields. But as this is very subjective, which order do you prefer?

thirdeyenick:
> > > Playing around with this locally, the "found in other project" suggestions work great. But I'm not a 100% sure about putting the `PROJECT` column first. I know this is also how kubectl does it but I'm just not totally sure that at a glance the first thing you want to see is the project instead of the resouce name:
> > > ```shell
> > > PROJECT    NAME                  HOSTS                              UNVERIFIED_HOSTS
> > > foo        docs-page             docs-page.edcd8e7.deploio.app      none
> > > foo        go-example            go-example.edcd8e7.deploio.app     none
> > > # vs
> > > NAME          PROJECT            HOSTS                              UNVERIFIED_HOSTS
> > > docs-page     foo                docs-page.edcd8e7.deploio.app      none
> > > go-example    foo                go-example.edcd8e7.deploio.app     none
> > > ```
> > 
> > 
> > As the display function also is kind of generic, the best option we have is to either add the project to the most left or most right position. We could implement some logic which shows the `PROJECT` column always to the right of a possibly given `NAME` column, but I am not sure if it is worth the effort (and magic) :-)
> 
> Ah yeah, unfortunately the `NAME` column is always added manually in the specific get implementations. We should have just passed the whole `client.Object` to `get.writeTabRow`, then we would be way more flexible as we always have the name and namespace fields. But as this is very subjective, which order do you prefer?

This is very subjective and I have no strong opinion on this tbh. Both is fine with me. I anyway read the column names.

ctrox (approved):
Let's leave it as is right now (column order), IMO we could always change it in the future as the standard output should not really be used in automation, we have `-o yaml` for that.

PR #162, opened 2024-09-17
fix displaying of nctl version

thirdeyenick:
This fixes the `nctl --version` command which was showing `dev` for released versions before.

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"runtime/debug"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -50,10 +50,15 @@ type rootCommand struct {
...
 
 const (
-	version           = "dev"
 	defaultAPICluster = "nineapis.ch"
 )
 
+var (
+	// this has to be a variable as it will be replaced by CI/CD
+	// (goreleaser)
+	version = "dev"
```
thde, line 59: I like to use this pattern: https://github.com/thde/truenas-scale-acme/blob/main/main.go#L53

It can still be set by the CI, but falls back to the Go native way if not.

thirdeyenick: Thanks, will have a look!

thirdeyenick: Added it. Could you have another look please?

gajicdev (approved):

Issue #160, opened 2024-09-16
nctl --version just shows `dev`


rnestler:
Since 1.7.4 `nctl --version` just shows `dev`:
```
% nctl --version
dev
```

I guess the reason is the change in https://github.com/ninech/nctl/commit/587c14bd92d38e61324c1cdca9ee5ebdb28605ac which changes the `version` from `var` to `const` and thus it's not overridable at link time anymore.

thirdeyenick:
This issue should be fixed in https://github.com/ninech/nctl/pull/162. So with the next released version this issue should not come up again.

thirdeyenick:
We just released `nctl` [v1.7.5](https://github.com/ninech/nctl/releases/tag/v1.7.5) which fixed this issue.

rnestler:
It now shows `1.7.5-arch (commit: 11465dedde74d0764d2962f5716f31b5d68bbf82, date: 2024-09-16T15:06:16Z)` on my system which is the wrong commit, since I build from the source tar ball inside a git repository, so it picks up that git commit from there.

thde:
Yes, that should be fixed by https://github.com/ninech/nctl/pull/165

thirdeyenick:
Might this issue be caused by some specific Arch Linux build steps? I can't find the commit `11465dedde74d0764d2962f5716f31b5d68bbf82` in the repository. Go does VCS stamping by default when building. That is where the output information comes from.

thirdeyenick:
When executing the build steps contained in the repository `aur.archlinux.org/nctl.git` I end up with:

```
$> ./nctl --version
v1.7.5-arch (commit: , date: )
```

I am not sure where the commit from the initial description comes from, but we should maybe not show empty commits/dates in the version output.

thirdeyenick:
Sorry, I now read your [comment](https://github.com/ninech/nctl/issues/160#issuecomment-2358844823) and understand that you build in a git repository. I think you could just use `go build -buildvcs=false ...` when building for Arch Linux to disable VCS stamping. On the other hand, we need to remove the empty `commit` and `date` output.

rnestler:
> Sorry, I now read your https://github.com/ninech/nctl/issues/160#issuecomment-2358844823 and understand that you build in a git repository.

Yes, AUR packages are just git repositories, so it picks up the git commit from the AUR package locally.

[...]

PR #165, opened 2024-09-23
build: add missing build variables

thde:


ctrox (approved):
I think this won't directly fix the issue mentioned in https://github.com/ninech/nctl/issues/160#issuecomment-2358844823 but it's good to set this explicitly anyways.

rnestler:
> I think this won't directly fix the issue mentioned in https://github.com/ninech/nctl/issues/160#issuecomment-2358844823 but it's good to set this explicitly anyways.

I can fix it the same way in the source build, by passing a commit hash.

thde:
> I think this won't directly fix the issue mentioned in https://github.com/ninech/nctl/issues/160#issuecomment-2358844823 but it's good to set this explicitly anyways.

Correct it wont. But it makes it more obvious what variables can be defined.

PR #171, opened 2024-10-01
Fix resource listing

thirdeyenick:
This fixes the not working resource listing over all projects. We now search for the users projects and search in them for the corresponding resource if `--all-projects/-A` was given. As this required quite some changes in test setups (projects have to be around when testing the `--all-projects` functionality), this PR also extends the internal test framework which can be used to setup a fake clientset. All manual creations of fake clientsets have been replaced by the test framework.

The main changes of this PR are in `internal/test/client.go` and `get/get.go`. All the other changes are test framework adaptions or cleanups of existing tests.


ctrox (approved):
One negative side-effect that I noticed is that now Niners with full permissions can't list objects in all namespaces anymore. But I think we can't really have both so we might just need to live with it.

thirdeyenick:
> One negative side-effect that I noticed is that now Niners with full permissions can't list objects in all namespaces anymore. But I think we can't really have both so we might just need to live with it.

Yeah, true this a negative side effect. This could be changed by adding an option, but I'd say it is rather uncommon to have something like an 'admin' option in such a tool.

api/config/extension_config.go
```diff
@@ -1,54 +1,43 @@
...
-		err,
-		format.Command().Login(),
-	)
-}
-
-// Config is used to store information in the kubeconfig context created
-type Config struct {
+// ExtensionConfig is used to store information in the kubeconfig context
+// created
+type ExtensionConfig struct {
```
ctrox, line 40: Nit: the name `config.ExtensionConfig` stutters a bit.

thirdeyenick: renamed it to `config.Extension` and adapted the other related functions

PR #174, opened 2024-10-08
fix: print warning when deletion of kubeconfig fails

ctrox:
Instead of erroring out when the removal of the vcluster from the kubeconfig fails, we instead just print a warning. The vcluster might have been created outside of nctl and might not be present in the local kubeconfig.


thirdeyenick (approved):
👍

PR #176, opened 2024-10-10
fix: release sort order

ctrox:
The OrderReleaseList returns the latest release last, which is not what the comment on the function says. Because of this, exec sometimes fails if there are two releases available (can happen when switching to a new release).

As a side effect, this also changes the order of the `get release` command. Should we reverse it for `get`?


thirdeyenick:
> The OrderReleaseList returns the latest release last, which is not what the comment on the function says. Because of this, exec sometimes fails if there are two releases available (can happen when switching to a new release).
> 
> As a side effect, this also changes the order of the `get release` command. Should we reverse it for `get`?

Good catch! I think reversing the order for `get release` might make sense as one would expect the latest release to be on the bottom?

ctrox:
> Good catch! I think reversing the order for `get release` might make sense as one would expect the latest release to be on the bottom?

Yeah I was thinking the same, especially since all other resources are sorted by the name, which for example makes the builds be sorted by "latest on the bottom" as they are numbered. I added a bool to reverse (just) the timestamp order, does this make sense to you?

thirdeyenick (approved):
👍

thirdeyenick:
> > Good catch! I think reversing the order for `get release` might make sense as one would expect the latest release to be on the bottom?
> 
> Yeah I was thinking the same, especially since all other resources are sorted by the name, which for example makes the builds be sorted by "latest on the bottom" as they are numbered. I added a bool to reverse (just) the timestamp order, does this make sense to you?

Yeah this makes sense.

PR #180, opened 2024-10-22
 feat: add worker jobs

ctrox:
This adds worker jobs to the app create/update commands. As a limitation
of CLI flags, it only allows to define one worker job on creation. With
update, you can add/remove as many as possible.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/moby/moby v27.1.1+incompatible
 	github.com/moby/term v0.5.0
-	github.com/ninech/apis v0.0.0-20241014111010-322e555390a0
+	github.com/ninech/apis v0.0.0-20241023072837-ff94192f11d1
 	github.com/posener/complete v1.2.3
 	github.com/prometheus/common v0.55.0
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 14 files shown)


go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20
 	github.com/moby/moby v27.1.1+incompatible
 	github.com/moby/term v0.5.0
-	github.com/ninech/apis v0.0.0-20241014111010-322e555390a0
+	github.com/ninech/apis v0.0.0-20241022115306-047cdfbe076b
```
ctrox, line 29: these changes depend on unmerged API changes so this will need to be updated again once the API is merged.

update/application.go
```diff
@@ -81,6 +84,12 @@ type deployJob struct {
 	Timeout *time.Duration `help:"Timeout of the job." placeholder:"${app_default_deploy_job_timeout}"`
 }
 
+type workerJob struct {
+	Name    *string `help:"Name of the worker job to add." placeholder:"worker-1"`
```
thirdeyenick, line 88: Do all the fields need to be `*string`? Wouldn't non pointer strings be sufficient and easier to handle?

thirdeyenick: Ok, I understand. 👍

thirdeyenick (approved):
lgtm 👍

PR #181, opened 2024-10-23
Description for filename parameter.

SaadAssaf:


create/create.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 type Cmd struct {
-	Filename            string               `short:"f" predictor:"file"`
+	Filename            string               `short:"f" help:"Specify the file containing the application definition." predictor:"file"`
```
ctrox, line 23: This isn't specific to applications, `nctl create -f` can be used for any resource. I write something similar to the `nctl apply -f` help: `Apply any resource from a yaml or json file.`

create/create.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 type Cmd struct {
-	Filename            string               `short:"f" help:"Specify the file containing the application definition." predictor:"file"`
+	Filename            string               `short:"f" help:"Apply any resource from a yaml or json file." predictor:"file"`
```
ctrox, line 23:
```suggestion
	Filename            string               `short:"f" help:"Create any resource from a yaml or json file." predictor:"file"`
```

ctrox (approved):

PR #183, opened 2024-10-30
wrote if statement to simplfy the forbidden error

SaadAssaf:


main.go
```diff
@@ -154,7 +156,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	kongCtx.FatalIfErrorf(kongCtx.Run(ctx, client))
+	err = kongCtx.Run(ctx, client)
+	if err != nil {
+		if k8serrors.IsForbidden(err) {
+			fmt.Println("nctl: Permission denied: are you part of the org?	")
```
thirdeyenick, line 162: I think this will print out a message, but the exit code of `nctl` will be 0, which wouldn't be correct I think. Maybe you could use another function of `kongCtx` to exit and print the generic permission error?

main.go
```diff
@@ -154,7 +156,16 @@ func main() {
...
+			
+		}else{
+			kongCtx.FatalIfErrorf(kongCtx.Run(ctx, client))
```
thirdeyenick, line 165: This would run the same command again (we already execute it at https://github.com/ninech/nctl/pull/183/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R159). We should just exit here with the error.

main.go
```diff
@@ -154,7 +156,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	kongCtx.FatalIfErrorf(kongCtx.Run(ctx, client))
+	err = kongCtx.Run(ctx, client)
+	if err != nil {
+		if k8serrors.IsForbidden(err) {
+			kongCtx.Fatalf("Permission denied: are you part of the org?")
```
thirdeyenick, line 162: one minor improvement suggestion:

```go
if err := kongCtx.Run(ctx, client); err != nil {
    if k8serrors.IsForbidden(err) {
        err = errors.New("got permission denied error: are you part of the organization?")
    }
    kongCtx.FatalIfErrorf(err)
}
```

One thing though which might be a bit problematic: we loose the original permission error. Maybe we can think about some `--debug` flag which would show the original error in addition?

main.go
```diff
@@ -34,6 +36,7 @@ type flags struct {
 	APICluster     string           `help:"Context name of the API cluster." default:"${api_cluster}" env:"NCTL_API_CLUSTER"`
 	LogAPIAddress  string           `help:"Address of the deplo.io logging API server." default:"https://logs.deplo.io" env:"NCTL_LOG_ADDR"`
 	LogAPIInsecure bool             `help:"Don't verify TLS connection to the logging API server." hidden:"" default:"false" env:"NCTL_LOG_INSECURE"`
+	Verbose        bool 			`help:"Show detailed error message."`
```
thirdeyenick, line 39:
```suggestion
	Verbose        bool 			`help:"Show verbose messages."`
```

main.go
```diff
@@ -154,7 +157,15 @@ func main() {
 		os.Exit(1)
 	}
 
-	kongCtx.FatalIfErrorf(kongCtx.Run(ctx, client))
+	err = kongCtx.Run(ctx, client)
+	if err != nil {
+		if k8serrors.IsForbidden(err) && !nctl.Verbose {
+			kongCtx.Fatalf("Permission denied: are you part of the organization?\nuse --verbose to see detailed error")
```
thirdeyenick, line 163: I'd do a

```go
if k8serrors.IsForbidden(err) && !nctl.Verbose {
err = errors.New("Permission denied: are you part of the organization?")
}
kongCtx.FatalIfErrorf(err)
```

I'd do this because you don't see the `else` keywork in golang code that often. You can also check https://go.dev/wiki/CodeReviewComments#indent-error-flow for an explanation of this.

main.go
```diff
@@ -154,7 +157,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	kongCtx.FatalIfErrorf(kongCtx.Run(ctx, client))
+	err = kongCtx.Run(ctx, client)
+	if err != nil {
+		if k8serrors.IsForbidden(err) && !nctl.Verbose {
+			err = errors.New("Permission denied: are you part of the organization?")
```
thirdeyenick, line 163:
```suggestion
			err = errors.New("permission denied: are you part of the organization?")
```

This should fix the linter error.

PR #190, opened 2024-11-25
feat: add hidden --all-namespaces flag

ctrox:
To be able to again list resources across all namespaces for debugging. The flag has been hidden so users should not stumble upon the flag unless they read the code.


thirdeyenick (approved):
👍

PR #194, opened 2024-12-05
feat: warn when setting project/org

ctrox:
When setting the project or organization, it will now warn you if the
project does not exist, your current user does not have access to it or
if the organization is not in the list of available organizations. We
don't error out to still support some use-cases where a user might not
have the permission to get a project or might not be fully part of an
org but might still have access to some resource in it.

The first commit just refactors some code I had to touch for this.

### DIFF
--- a/auth/set_project.go
+++ b/auth/set_project.go
@@ -2,9 +2,14 @@ package auth
 
 import (
 	"context"
+	"fmt"
 
+	management "github.com/ninech/apis/management/v1alpha1"
 	"github.com/ninech/nctl/api"
 	"github.com/ninech/nctl/api/config"
+	"github.com/ninech/nctl/internal/format"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 type SetProjectCmd struct {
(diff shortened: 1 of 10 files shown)


auth/set_project.go
```diff
@@ -2,15 +2,44 @@ package auth
...
+	}
+
+	// we get the project without using the result to be sure it exists and the
+	// user has access.
+	if err := client.Get(ctx, types.NamespacedName{Name: s.Name, Namespace: org}, &management.Project{}); err != nil {
+		if !errors.IsNotFound(err) && !errors.IsForbidden(err) {
+			return err
+		}
+		if errors.IsNotFound(err) {
+			format.PrintWarningf("did not find Project %s in your Organization %s.\n", s.Name, org)
```
pawelkuc, line 32: We don't throw an error here - that's fine.

But I just wonder if we're not making things more confusing. For example: in this situation I can set a non-existent project "idonotexist". Then I can for example ask for releases in a separate `nctl get releases` query and I get: `no Releases found in project idontexist` (I could have forgotten that I had this warning that this project doesn't exist). What I'm trying to say is that it would be nice if we could have this "warning" somehow persistent throughout the nctl context. But that won't be so easy to change, I think.

A similar problem is with set org - you can just change the organization in your kubeconfig and `nctl auth whoami` will just read it without warning:

```
You are currently logged in the with the following account: "pawel@nine.ch"
Your current organization: "supersecretnonexistentorganization"

Available Organizations:
	nine

To switch the organization use the following command:
$ nctl auth set-org <org-name>
```

and `nctl get releases` will just report than no releases found (project/namespace name doesn't matter here):
```
no Releases found in project <projectname>
```

pawelkuc (approved):

auth/set_project.go
ctrox: One problem is projects are namespaced and live in the org namespace. So in it could be that a project exists in another organization so if we would throw an error, the user would first need to switch the org before being able to switch the project.

> What I'm trying to say is that it would be nice if we could have this "warning" somehow persistent throughout the nctl context. But that won't be so easy to change, I think.

Yeah I think that would be pretty hard to achieve. We could make a `Get` to the project before every other call we make but then we'd have a similar problem where you could be part of two organizations and you make a `nctl get app x -p bar` while having the org set to `foo`. This would then error even though we are also part of the `bar` org and are allowed to get that app.

pawelkuc: that's true. I guess we don't change anything now since your solution works. 
But maybe we'll come back to this discussion at some point in the future 🤷‍♂️ 😄

ctrox: Yeah I'm pretty sure it will be discussed again at some point :smile:

PR #197, opened 2024-12-13
auth login in non-interactive env should not block

pawelkuc:
If you execute `nctl auth login` in a non-interactive environment (e.g. CI/CD) and forget to set `--api-token`, the browser should not try to open - you should see an error message instead.

### DIFF
--- a/auth/login.go
+++ b/auth/login.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


auth/login.go
```diff
@@ -24,6 +25,8 @@ type LoginCmd struct {
 	ClientID     string `help:"Client ID is the OIDC client ID of the API." default:"nineapis.ch-f178254"`
 }
 
+const ErrNonInteractiveEnvironmentEptyToken = "a static API token is required in non-interactive envirtonments"
```
thirdeyenick, line 28:
```suggestion
const ErrNonInteractiveEnvironmentEmptyToken = "a static API token is required in non-interactive environments"
```

thirdeyenick (commented):
I am not sure why we have to add this specific option to the context and then have to change so many function calls (just because of that one option). For example, why do we need that interactive env information in the `format.getPrinter` function? Maybe you could help me in understanding this?

auth/login.go
thirdeyenick: I had a look and it seems we want to prevent an error in the `TestLoginCmd` test. Instead of using the context for this, could we also just an env variable or a hidden field on the `LoginCmd` struct? Something with which we can force the usage of the tokenGetter and prevent the error.

pawelkuc: >  why do we need that interactive env information in the format.getPrinter function? Maybe you could help me in understanding this?

The overall question was: how I can control the `func IsInteractiveEnvironment()`

The first idea was to use a specific environment variables when running tests, such as GO_TESTING=true.
You can then check for these variables to determine if the call is from a test.
Cons: tied directly to the Go testing environment via environment variables.

Next idea was to add an additional argument or use a package-level variable that can be toggled in your tests to force the function to return true. 
This gives my a fine-grained control over when interactivity is simulated.

But the package-level variable or env variable introduces potential race conditions when tests are run in parallel,
as parallel tests may set or unset the variable concurrently.

So the next idea was to introduce a `context.Context` parameter to propagate this `forceInteractiveEnvironment` "flag" without relying on global variables.
This avoids race conditions entirely and is more idiomatic in Go. 
Of course, this also means passing context to the `IsInteractiveEnvironment()` function.

We already use context e.g. in `cmd.printStats(ctx, client, appList.Items, get, defaultOut(cmd.out))`, so I think we are quite consistent by introducing context to `format.PrettyPrintObjects()` as well. 

And `format.getPrinter` already had the logic for detecting interactive env:
```
...
	f, isFile := out.(*os.File)
	if !isFile {
		return p, nil
	}
	if isatty.IsTerminal(f.Fd()) || isatty.IsCygwinTerminal(f.Fd()) {
...
```

I just created a separate func `func IsInteractiveEnvironment()` to make it reusable.

Does that better explain the change? :)

thirdeyenick: Hmm, but cant' we add a hidden field to the `LoginCmd` struct? We can then set this field in the specific test and overcome parallel running test issues. I assume the `getPrinter` function was working before as we mostly use a buffer for output in tests? So I am not sure we need the "forceTerminalUsage" there as well.

pawelkuc: sure - hidden field in `LoginCmd` is also a good option. I will prepare this change.

auth/login.go
```diff
@@ -58,6 +62,12 @@ func (l *LoginCmd) Run(ctx context.Context, command string, tk api.TokenGetter)
 		return login(ctx, cfg, loadingRules.GetDefaultFilename(), userInfo.User, "", project(l.Organization))
 	}
 
+	if !l.ForceInteractiveEnvOverride {
```
thirdeyenick, line 65:
```suggestion
	if !l.ForceInteractiveEnvOverride && !format.IsInteractiveEnvironment(os.Stdout) {
```

auth/login.go
thirdeyenick: Looks good  👍

thirdeyenick (approved):

PR #206, opened 2025-01-21
only show credentials if basic-auth was set by user

thirdeyenick:
As the basic-auth secret is now handled separately from the basic-auth settings (enabled/disabled), it will be constantly referenced in the application status. That is why we can not use it anymore to decide if basic-auth was enabled or not.

### DIFF
--- a/create/application.go
+++ b/create/application.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ninech/nctl/logs"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


ctrox:
Hmm, what if basic auth is enabled through `deploio.yaml`? Shouldn't we use the release to determine if it's enabled/disabled?

create/application.go
```diff
@@ -514,6 +520,23 @@ func waitForRelease(app *apps.Application) waitStage {
 		},
 	}
 }
+func latestReleaseHasBasicAuthEnabled(ctx context.Context, c runtimeclient.Reader, app *apps.Application) (bool, error) {
```
ctrox, line 523: Nit: newline before the new function starts

PR #208, opened 2025-02-07
Improve Logs UX

ctrox:
* Increases default look-back to match retention
* Add error messages when requested logs exceed retention
* Adds flags to get logs for absolute time ranges
* Add error message for when no logs have been found

### DIFF
--- a/logs/build.go
+++ b/logs/build.go
@@ -3,6 +3,7 @@ package logs
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	apps "github.com/ninech/apis/apps/v1alpha1"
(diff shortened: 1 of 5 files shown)


logs/build.go
```diff
@@ -24,6 +25,12 @@ func (cmd *buildCmd) Run(ctx context.Context, client *api.Client) error {
 		if err := client.Get(ctx, api.NamespacedName(cmd.Name, client.Project), build); err != nil {
 			return err
 		}
+		if cmd.Since > 0 && time.Since(build.CreationTimestamp.Time) > cmd.Since {
```
thirdeyenick, line 28: Not sure I fully understand, but this path is only taken if the user specified logs for a specific build, right? Shouldn't the since parameter then be ignored as the user clearly indicated that he wants the logs of a specific build?

ctrox: Yeah I changed this to be less confusing. Instead of using `cmd.Since` (which defaults to `logRetention` if not set) I just use `logRetention` directly. This is just so the error makes a bit more sense for the builds, without this, users would just get the generic `the logs requested exceed the retention period` error.

logs/logs.go
```diff
@@ -24,18 +24,31 @@ type resourceCmd struct {
...
-	out      output.LogOutput
+	out      log.Output
 }
 
+var logRetention = time.Duration(time.Hour * 24 * 30)
+
 func (cmd *logsCmd) Run(ctx context.Context, client *api.Client, queryString string, labels ...string) error {
+	now := time.Now()
+	start, end := now.Add(-cmd.Since), now
+	if !cmd.From.IsZero() && !cmd.To.IsZero() {
```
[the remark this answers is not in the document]
ctrox: Right, I separated these two conditions so `From` can be set individually.

thirdeyenick (approved):
👍

PR #210, opened 2025-02-07
refactor: move list to api package

ctrox:
This makes our special list function available to other packages than get and uses it in logs.

### DIFF
--- a/api/list.go
+++ b/api/list.go
@@ -1,0 +1,216 @@
+package api
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"slices"
+	"strings"
+
+	management "github.com/ninech/apis/management/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/conversion"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ListOpts struct {
+	clientListOptions []runtimeclient.ListOption
+	searchForName     string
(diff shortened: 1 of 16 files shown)


api/list.go
```diff
@@ -0,0 +1,214 @@
...
+				"Maybe you want to use the '--project' flag to specify one of these projects?",
+			strings.Join(foundInProjects, " ,"),
+		)
+	}
+	return errors.New(errorMessage)
+}
+
+// ListN lists resources with some ux-improvements like hinting when a resource
+// has been found in a different project of the same organization.
+func (c *Client) ListN(ctx context.Context, list runtimeclient.ObjectList, options ...ListOpt) error {
```
ctrox, line 72: Name is a placeholder, open to suggestions. I also thought about simply shadowing the client list function but I don't think that's a good idea as I still want to be able to use the direct list.

gajicdev: You are right about the shadowing, we might want to still use direct list somewhere.

I like the name, the only other one I could come up is `ListResources` or something like that, but I like the ListN name

gajicdev (approved):

api/list.go
ctrox: I now changed the name to simply `ListObjects` (and also added `GetObjects`). Makes sense to me as it takes an `ObjectList`. WDYT?

gajicdev (approved):

api/list.go
gajicdev: great name, approved 👍

PR #212, opened 2025-03-09
add scheduled jobs

gajicdev:


update/application.go
```diff
@@ -336,6 +351,55 @@ func deleteWorkerJob(name string, cfg *apps.Config) {
 	cfg.WorkerJobs = newJobs
 }
 
+func (job scheduledJob) applyUpdates(cfg *apps.Config) {
+	if (job.Command != nil || job.Size != nil) && job.Name == nil {
```
thirdeyenick, line 355: I think a `job.Name` is needed all the time, right? So a first check for

```
if job.Name == nil {
  return
}
```

should be done.

update/application.go
```diff
@@ -336,6 +351,55 @@ func deleteWorkerJob(name string, cfg *apps.Config) {
...
+		format.PrintWarningf("you need to pass a job name to update the command or size\n")
+	}
+	if job.Name == nil {
+		return
+	}
+	found := false
```
thirdeyenick, line 361: Not sure we need this. Once you found a job to update (first range loop) you can just return at the end.

gajicdev: I got this from here: https://github.com/ninech/nctl/blob/main/update/application.go#L293

Should I also change it there or do we keep em the same?

thirdeyenick: Maybe you find a problem with the suggested approach, but if not I'd also update it there.

create/application.go
```diff
@@ -75,6 +76,13 @@ type workerJob struct {
 	Size    *string `help:"Size of the worker (defaults to \"${app_default_size}\")." placeholder:"${app_default_size}"`
 }
 
+type scheduledJob struct {
+	Command  string  `help:"Command to execute to start the scheduled job." placeholder:"\"bundle exec sidekiq\""`
```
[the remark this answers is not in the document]
gajicdev: changed to bundle exec rails runner wdyt?

create/application.go
```diff
@@ -75,6 +76,13 @@ type workerJob struct {
...
+	Name     string  `help:"Name of the scheduled job job to add." placeholder:"scheduled-1"`
+	Size     *string `help:"Size of the scheduled (defaults to \"${app_default_size}\")." placeholder:"${app_default_size}"`
```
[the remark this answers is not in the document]
gajicdev: added

update/application.go
```diff
@@ -91,6 +93,13 @@ type workerJob struct {
 	Size    *string `help:"Size of the worker (defaults to \"${app_default_size}\")." placeholder:"${app_default_size}"`
 }
 
+type scheduledJob struct {
+	Command  *string `help:"Command to execute to start the scheduled job." placeholder:"\"bundle exec sidekiq\""`
```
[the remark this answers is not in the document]
gajicdev: same as above, using bundle exec rails runner

update/application.go
```diff
@@ -91,6 +93,13 @@ type workerJob struct {
...
+	Name     *string `help:"Name of the scheduled job job to add." placeholder:"scheduled-1"`
+	Size     *string `help:"Size of the scheduled (defaults to \"${app_default_size}\")." placeholder:"${app_default_size}"`
```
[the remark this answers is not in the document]
gajicdev: added

update/application.go
gajicdev: I like your approach better and added it for both functions

create/application.go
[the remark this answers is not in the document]
thirdeyenick: yeah, seems like to be a valid one off command in the rails world.

create/application.go
```diff
@@ -298,6 +306,21 @@ func (app *applicationCmd) config() apps.Config {
 		config.WorkerJobs = append(config.WorkerJobs, workerJob)
 	}
 
+	for _, job := range app.ScheduledJob {
+		if len(job.Command) != 0 && len(job.Name) != 0 && len(job.Schedule) != 0 {
+			scheduledJob := apps.ScheduledJob{
+				Job: apps.Job{
+					Name:    job.Name,
+					Command: job.Command,
```
thirdeyenick, line 314: not sure, but isn't the schedule missing here?

update/application.go
```diff
@@ -336,6 +348,56 @@ func deleteWorkerJob(name string, cfg *apps.Config) {
 	cfg.WorkerJobs = newJobs
 }
 
+func (job scheduledJob) applyUpdates(cfg *apps.Config) {
+	if job.Name == nil {
+		return
+	}
+
+	if job.Command != nil || job.Size != nil {
+		format.PrintWarningf("you need to pass a job name to update the command or size\n")
```
thirdeyenick, line 357: I'd say this warning message should go into the aboves `if` clause

update/application.go
```diff
@@ -336,6 +348,56 @@ func deleteWorkerJob(name string, cfg *apps.Config) {
...
+	}
+	if job.Name == nil {
```
thirdeyenick, line 359: this one can be removed I'd say?

update/application.go
```diff
@@ -336,6 +348,56 @@ func deleteWorkerJob(name string, cfg *apps.Config) {
 	cfg.WorkerJobs = newJobs
 }
 
+func (job scheduledJob) applyUpdates(cfg *apps.Config) {
+	if job.Name == nil {
+		return
+	}
+
+	if job.Command != nil || job.Size != nil {
```
thirdeyenick, line 356: not sure this still needs to happen. We check this again further down in the loop.

thirdeyenick (approved):

PR #213, opened 2025-03-14
fix stats displaying

thirdeyenick:
This improves statistics displaying for app replicas, worker job replicas and scheduled job replicas by considering the individual max sizes

### DIFF
--- a/get/application.go
+++ b/get/application.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	metricsv1beta1 "k8s.io/metrics/pkg/apis/metrics/v1beta1"
+	"k8s.io/utils/ptr"
 )
 
 type applicationsCmd struct {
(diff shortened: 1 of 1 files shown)


get/application.go
```diff
@@ -200,55 +225,96 @@ func (cmd *applicationsCmd) printStats(ctx context.Context, c *api.Client, appLi
 		return err
 	}
 	w := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
-	get.writeHeader(w, "NAME", "REPLICA", "STATUS", "CPU", "CPU%", "MEMORY", "MEMORY%", "RESTARTS", "LASTEXITCODE")
+	get.writeHeader(w, "NAME", "TYPE", "REPLICA", "STATUS", "CPU", "CPU%", "MEMORY", "MEMORY%", "RESTARTS", "LASTEXITCODE")
```
ctrox, line 228: Isn't the type already visible from the replica name? Just not sure if we should really make the stats output even wider since it's already super wide if you have a long app name.

ctrox (approved):

PR #216, opened 2025-03-21
Parallelize fetching resource for all projects

rnestler:
This improves fetching all applications for our organization by a factor of ~14:
```
% hyperfine --warmup 1 "./nctl get applications -A" "nctl get applications -A"
Benchmark 1: ./nctl get applications -A
  Time (mean ± σ):     243.6 ms ±  27.1 ms    [User: 128.8 ms, System: 59.8 ms]
  Range (min … max):   216.0 ms … 292.8 ms    11 runs

Benchmark 2: nctl get applications -A
  Time (mean ± σ):      3.389 s ±  0.012 s    [User: 0.144 s, System: 0.067 s]
  Range (min … max):    3.374 s …  3.409 s    10 runs

Summary
  ./nctl get applications -A ran
   13.91 ± 1.55 times faster than nctl get applications -A
```

The same holds for getting all `postgres` databases:
```
% hyperfine --warmup 1 "./nctl get postgres -A" "nctl get postgres -A"
Benchmark 1: ./nctl get postgres -A
  Time (mean ± σ):     248.7 ms ±  30.3 ms    [User: 158.9 ms, System: 61.3 ms]
  Range (min … max):   213.0 ms … 300.5 ms    10 runs

Benchmark 2: nctl get postgres -A
  Time (mean ± σ):      3.433 s ±  0.064 s    [User: 0.150 s, System: 0.060 s]
  Range (min … max):    3.389 s …  3.602 s    10 runs

Summary
  ./nctl get postgres -A ran
   13.81 ± 1.70 times faster than nctl get postgres -A
```

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -242,6 +242,8 @@ func loadConfigWithContext(apiServerURL string, loader clientcmd.ClientConfigLoa
 	}
 
 	cfg, err := clientConfig.ClientConfig()
+	cfg.QPS = 50
+	cfg.Burst = 100
 	return cfg, ns, err
 }
 
--- a/api/client.go
+++ b/api/client.go
@@ -242,8 +242,8 @@ func loadConfigWithContext(apiServerURL string, loader clientcmd.ClientConfigLoa
 	}
 
 	cfg, err := clientConfig.ClientConfig()
-	cfg.QPS = 50
-	cfg.Burst = 100
+	cfg.QPS = 25
+	cfg.Burst = 50
 	return cfg, ns, err
 }
 
--- a/api/list.go
+++ b/api/list.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"log"
@@ -127,10 +128,15 @@ func (c *Client) ListObjects(ctx context.Context, list runtimeclient.ObjectList,
 		return fmt.Errorf("error when searching for projects: %w", err)
 	}
 
+	type ProjectItems struct {
+		projectName string
+		items       reflect.Value
+	}
+
 	projectsSize := len(projects)
 	var wg sync.WaitGroup
 	wg.Add(projectsSize)
-	ch := make(chan reflect.Value, projectsSize)
+	ch := make(chan ProjectItems, projectsSize)
 
 	for _, proj := range projects {
 		tempOpts := slices.Clone(opts.clientListOptions)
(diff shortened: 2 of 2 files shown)


api/list.go
```diff
@@ -125,19 +128,35 @@ func (c *Client) ListObjects(ctx context.Context, list runtimeclient.ObjectList,
...
-			items.Set(reflect.Append(items, tempListItems.Index(i)))
+		go func() {
+			defer wg.Done()
+			// we ensured the list is a pointer type and that is has an
+			// 'Items' field which is a slice above, so we don't need to do
+			// this again here and instead use the reflect functions directly.
+			tempList := reflect.New(reflect.TypeOf(list).Elem()).Interface().(runtimeclient.ObjectList)
+			tempList.GetObjectKind().SetGroupVersionKind(list.GetObjectKind().GroupVersionKind())
+			if err := c.List(ctx, tempList, append(tempOpts, runtimeclient.InNamespace(proj.Name))...); err != nil {
+				log.Printf("error when searching in project %s: %w", proj.Name, err)
```
rnestler, line 146: I guess just logging the error is good enough

api/client.go
```diff
@@ -242,6 +242,8 @@ func loadConfigWithContext(apiServerURL string, loader clientcmd.ClientConfigLoa
 	}
 
 	cfg, err := clientConfig.ClientConfig()
+	cfg.QPS = 50
+	cfg.Burst = 100
```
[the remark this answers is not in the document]
rnestler: Let's test with 25 / 50.

api/list.go
```diff
@@ -125,19 +128,35 @@ func (c *Client) ListObjects(ctx context.Context, list runtimeclient.ObjectList,
 		return fmt.Errorf("error when searching for projects: %w", err)
 	}
 
+	projectsSize := len(projects)
+	var wg sync.WaitGroup
+	wg.Add(projectsSize)
+	ch := make(chan reflect.Value, projectsSize)
+
```
rnestler, line 135: Maybe limit the number of concurrent go routines as well.

api/list.go
ctrox: I would use our existing printing function here `format.PrintWarningf` as `log.Printf` won't really be consistent with other nctl output.

api/client.go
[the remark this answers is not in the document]
rnestler: 25 / 50 works fine for Renuo's projects.

api/list.go
rnestler: Fixed.

api/list.go
```diff
@@ -125,19 +128,35 @@ func (c *Client) ListObjects(ctx context.Context, list runtimeclient.ObjectList,
...
+			// 'Items' field which is a slice above, so we don't need to do
+			// this again here and instead use the reflect functions directly.
+			tempList := reflect.New(reflect.TypeOf(list).Elem()).Interface().(runtimeclient.ObjectList)
+			tempList.GetObjectKind().SetGroupVersionKind(list.GetObjectKind().GroupVersionKind())
+			if err := c.List(ctx, tempList, append(tempOpts, runtimeclient.InNamespace(proj.Name))...); err != nil {
+				log.Printf("error when searching in project %s: %w", proj.Name, err)
+				return
+			}
+			tempListItems := reflect.ValueOf(tempList).Elem().FieldByName("Items")
+			ch <- tempListItems
```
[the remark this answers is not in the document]
rnestler: Fixed.

api/list.go
rnestler: I didn't limit it for now. Since we already have the rate limit throttling of the k8s client library it doesn't make much sense. We can easily have 100s if not 1000s of concurrently running go routines without any issues.

ctrox (approved):
One thing that can be a bit annoying is that if you do run into the rate limit (which is not too likely at the moment), [client-go outputs](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L683) a (rather ugly) warning message. I tried to suppress it by discarding `klog` output but I was unable to get it to work right now. It's a detail, we can revisit it later if it bothers us too much.

So LGTM, thanks!

PR #218, opened 2025-04-07
feat: remove app git auth inline fields

pawelkuc:
Deploio apps currently support two methods for Git authentication:

* `spec.forProvider.git.auth.username` and `password` for inline credentials
* `spec.forProvider.git.auth.fromSecret` to reference a Kubernetes secret

In practice, `nctl` exclusively use the `fromSecret` method. Maintaining both options adds unnecessary complexity, and it looks like the inline method is not used at the moment.


thirdeyenick (approved):
👍

PR #223, opened 2025-04-23
add pausing of apps

gajicdev:


get/application.go
```diff
@@ -84,7 +84,7 @@ func printApplication(apps []apps.Application, get *Cmd, out io.Writer, header b
 	w := tabwriter.NewWriter(out, 0, 0, 4, ' ', 0)
 
 	if header {
-		get.writeHeader(w, "NAME", "REPLICAS", "WORKERJOBS", "SCHEDULEDJOBS", "HOSTS", "UNVERIFIEDHOSTS")
+		get.writeHeader(w, "NAME", "REPLICAS", "WORKERJOBS", "SCHEDULEDJOBS", "HOSTS", "UNVERIFIEDHOSTS", "PAUSED")
```
ctrox, line 87: Should we really make this wider once again? For most apps this will just show `false` and if it's paused you should also see that you have 0 replicas/worker jobs.

ctrox (approved):

PR #224, opened 2025-05-06
refactor: CloudVM help

fluser:
* Add default information: ubuntu24.04
* Mention available OS flavors and versions


create/cloudvm.go
```diff
@@ -20,7 +20,7 @@ type cloudVMCmd struct {
 	MachineType         string            `default:"" help:"The machine type defines the sizing for a particular CloudVM."`
 	Hostname            string            `default:"" help:"Hostname allows to set the hostname explicitly. If unset, the name of the resource will be used as the hostname. This does not affect the DNS name."`
 	PowerState          string            `default:"on" help:"Specify the initial power state of the CloudVM. Set to off to create "`
-	OS                  string            `default:"" help:"OS which should be used to boot the VM."`
+	OS                  string            `default:"ubuntu24.04" help:"OS which should be used to boot the VM. Available options: ubuntu{24.04,22.04,20.04}, rocky9, debian12"`
```
thde, line 23: If we hard code the values, we'll also need to change the code every time we add or remove an OS. 

I agree it would be nice to show them here, but then we should do it by reading the defaults from the API package like with MySQL:
https://github.com/ninech/nctl/blob/main/create/mysql.go#L118

fluser: I'm not very familiar with Go, but I hope my latest changes meet the requirements :smile:

PR #226, opened 2025-05-13
Fix error message for missing build name

schmijos:


logs/build.go
```diff
@@ -18,7 +18,7 @@ type buildCmd struct {
 
 func (cmd *buildCmd) Run(ctx context.Context, client *api.Client) error {
 	if cmd.Name == "" && cmd.ApplicationName == "" {
```
schmijos, line 20: The check is futile as well, right?

```suggestion
	if cmd.Name == "" {
```

schmijos:
https://deploiocommunity.slack.com/archives/C05LUUQ3UPN/p1747140455164189?thread_ts=1747135952.766839&cid=C05LUUQ3UPN

PR #236, opened 2025-06-05
feat: add json output

SaadAssaf:


internal/format/print.go
```diff
@@ -185,22 +200,35 @@ func PrettyPrintObject(obj any, opts PrintOpts) error {
 // printResource prints the resource similar to how
 // https://github.com/goccy/go-yaml#ycat does it.
 func printResource(obj any, opts PrintOpts) error {
-	b, err := yaml.Marshal(obj)
+	var b []byte
+	var err error
+	if opts.Format == OutputFormatTypeJSON {
```
thirdeyenick, line 205: maybe we can work with `switch` statements here?

ctrox:
Just two things I observed when testing locally:

1. The json output is all on one line, this could easily be changed by using `json.MarshalIndent(obj, "", "  ")` (two spaces) instead of `json.Marshal`. WDYT?
2. There is no final line break making it look ugly in some terminals (promt gets printed on the same line) and it makes copy-pasting the output harder.

ctrox (approved):

PR #247, opened 2025-07-02
feat(serviceconnection): add ressource

olknegate:
This MR is adding support for [ServiceConnections](https://docs.nine.ch/api/#tag/ServiceConnection) to `nctl`.


create/serviceconnection.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/alecthomas/kong"
+	meta "github.com/ninech/apis/meta/v1alpha1"
+	networking "github.com/ninech/apis/networking/v1alpha1"
+	"github.com/ninech/nctl/api"
+)
+
+type serviceConnectionCmd struct {
+	resourceCmd
+	Source                   string                    `placeholder:"kubernetescluster/my-clustername" help:"The source of the connection in the form kind/name. Allowed source kinds are: ${allowed_sources}." required:""`
```
thde, line 19: What about adding a `Reference` type and implementing the `encoding.TextUnmarshaler` or `json.Unmarshaler`:

https://github.com/alecthomas/kong?tab=readme-ov-file#custom-decoders-mappers

This would make the code a bit cleaner IMO and easier to reuse.

create/serviceconnection.go
```diff
@@ -0,0 +1,213 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+
+	"github.com/alecthomas/kong"
+	meta "github.com/ninech/apis/meta/v1alpha1"
+	networking "github.com/ninech/apis/networking/v1alpha1"
+	"github.com/ninech/nctl/api"
+)
+
+type serviceConnectionCmd struct {
```
thde, line 17: This implements the resource the same way as in the API, which makes sense. But from a Usability POV it might be nice to update an App or KVS instead of having to create a separate service connection:

```bash
nctl update app web-prod --service-connections kvs/sessions-prod
```

or

```bash
nctl update kvs sessions-prod --service-connections app/web-prod
```

What do you think?

go.sum
```diff
@@ -539,6 +542,18 @@ github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f/go.mod h1:ZdcZmHo+
...
+github.com/ninech/apis v0.0.0-20250701145711-b8a1280e92b0 h1:q5X0N8Q7uCGKZF5kds2e598UR1EzEnfnZl7DEJUXzRk=
+github.com/ninech/apis v0.0.0-20250701145711-b8a1280e92b0/go.mod h1:6srtlYi3nj8GRxQkbhvSbanIBc9vsoPLFOp5VQzNNhM=
+github.com/ninech/apis v0.0.0-20250701160639-c511ecdd1fe4 h1:5ZNMyniBQo/mPaRjZGdI+H5Dy4v+0xg4U/286mKOz2E=
+github.com/ninech/apis v0.0.0-20250701160639-c511ecdd1fe4/go.mod h1:6srtlYi3nj8GRxQkbhvSbanIBc9vsoPLFOp5VQzNNhM=
+github.com/ninech/apis v0.0.0-20250702115637-35129797c48c h1:ofxJ3HQx8B5bSS0GiMT/L90P0+oXjVizj5VV3JF/wiI=
+github.com/ninech/apis v0.0.0-20250702115637-35129797c48c/go.mod h1:6srtlYi3nj8GRxQkbhvSbanIBc9vsoPLFOp5VQzNNhM=
+github.com/ninech/apis v0.0.0-20250702131856-edc51fc0c1b1 h1:cltEEz8L8FHR7wai1WkoLM/4rDbKthsRxa1xMJqUgD4=
+github.com/ninech/apis v0.0.0-20250702131856-edc51fc0c1b1/go.mod h1:6srtlYi3nj8GRxQkbhvSbanIBc9vsoPLFOp5VQzNNhM=
+github.com/ninech/apis v0.0.0-20250702143341-d8aa2ea682a4 h1:EnR0YMDuH8ENcKxQbMihvpyoaZ7TWyKYw9h7FQx4Wdk=
+github.com/ninech/apis v0.0.0-20250702143341-d8aa2ea682a4/go.mod h1:6srtlYi3nj8GRxQkbhvSbanIBc9vsoPLFOp5VQzNNhM=
```
pawelkuc, line 556: Having these extra lines does not hurt but maybe we can run `go mod tidy`?

create/serviceconnection_test.go
```diff
@@ -0,0 +1,258 @@
...
+			want: networking.ServiceConnectionParameters{
+				Source: networking.Source{
+					Reference: meta.TypedReference{
+						Reference: meta.Reference{
+							Name:      "test-application",
+							Namespace: "default",
+						},
+						GroupKind: metav1.GroupKind{
+							Group: "apps.nine.ch",
+							Kind:  "Application",
```
pawelkuc, line 49: instead of hardcoding we can take this from github.com/ninech/apis/apps/v1alpha1
(typical in this test)

update/serviceconnection_test.go
```diff
@@ -0,0 +1,151 @@
...
+			want: networking.ServiceConnectionParameters{
+				Source: networking.Source{
+					Reference: meta.TypedReference{
+						Reference: meta.Reference{
+							Name:      "test-cluster",
+							Namespace: "default",
+						},
+						GroupKind: metav1.GroupKind{
+							Group: "cluster.nine.ch",
+							Kind:  "KubernetesCluster",
```
pawelkuc, line 72: instead of hardcoding we can take this from github.com/ninech/apis/apps/v1alpha1
(typical in this test)
also: group should be `infrastructure.nine.ch` here

internal/test/serviceconnection.go
```diff
@@ -0,0 +1,42 @@
...
+			ForProvider: networking.ServiceConnectionParameters{
+				Source: networking.Source{
+					Reference: meta.TypedReference{
+						Reference: meta.Reference{
+							Name:      "test-cluster",
+							Namespace: "default",
+						},
+						GroupKind: metav1.GroupKind{
+							Group: "cluster.nine.ch",
+							Kind:  "KubernetesCluster",
```
pawelkuc, line 25: instead of hardcoding we can take this from github.com/ninech/apis/apps/v1alpha1
(typical in this file)
also: group should be `infrastructure.nine.ch` here

update/serviceconnection.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type KubernetesClusterOptions struct {
+	PodSelector       LabelSelector `embed:"" prefix:"pod-selector-"`
+	NamespaceSelector LabelSelector `embed:"" prefix:"namespace-selector-"`
+}
+
+type LabelSelector struct {
+	MatchLabels      map[string]string          `help:"Map of labels(key-value-pairs) to match."`
+	MatchExpressions []LabelSelectorRequirement `embed:"" prefix:"match-expressions-"`
```
pawelkuc, line 27: the same issue as in `create/serviceconnection.go` (embedding slice of structs)

update/serviceconnection.go
```diff
@@ -0,0 +1,95 @@
...
+			return fmt.Errorf("resource is of type %T, expected %T", current, networking.ServiceConnection{})
+		}
+
+		return cmd.applyUpdates(sc)
+	}).Update(ctx)
+}
+
+func (cmd *serviceConnectionCmd) applyUpdates(sc *networking.ServiceConnection) error {
+	kubernetesClusterOptions := cmd.kubernetesClusterOptions()
+	if kubernetesClusterOptions != nil {
```
pawelkuc, line 56: this will never be nil. see my comment `create/serviceconnection.go`

create/serviceconnection.go
```diff
@@ -0,0 +1,169 @@
...
+	requirements := []metav1.LabelSelectorRequirement{}
+
+	for _, expression := range expressions {
+		requirements = append(requirements, metav1.LabelSelectorRequirement(expression))
+	}
+
+	return requirements
+}
+
+func groupKindFromKind(kind string) (metav1.GroupKind, error) {
```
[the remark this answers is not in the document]
thde: The thing is that we use the kinds for different things: In here we try to lookup the `GroupVersionKind` based on the `GroupKind`. But in https://github.com/ninech/nctl/blob/dirk/add_serviceconnections/get/all.go#L178 we want to list all nine specific types. 

I don't think the performance implications are that big. The max here is O(2N) (we could bring it down to O(N), but you'll always need to iterate at least once. ) Or am I understanding something wrong?

pawelkuc (approved):

create/serviceconnection.go
```diff
@@ -149,19 +193,19 @@ func (cmd *serviceConnectionCmd) newServiceConnection(namespace string) (*networ
 	return sc, nil
 }
 
-func groupKindFromKind(kind string) (metav1.GroupKind, error) {
+func groupVersionKindFromKind(kind string) (schema.GroupVersionKind, error) {
```
thirdeyenick, line 196: Do we maybe want to check the `group` as well? Just if we will ever have the same `Kind` in multiple groups?

PR #248, opened 2025-07-04
feat(databases): add commands

9marco:
Implements ninech/managed-services/backlog#1007, ninech/managed-services/backlog#961

### DIFF
--- a/get/get.go
+++ b/get/get.go
@@ -24,6 +24,7 @@ type Cmd struct {
 	Releases            releasesCmd           `cmd:"" group:"deplo.io" name:"releases" aliases:"release" help:"Get deplo.io Releases."`
 	Configs             configsCmd            `cmd:"" group:"deplo.io" name:"configs" aliases:"config" help:"Get deplo.io Project Configuration."`
 	MySQL               mySQLCmd              `cmd:"" group:"storage.nine.ch" name:"mysql" help:"Get MySQL instances."`
+	MySQLDatabases      mysqlDatabaseCmd      `cmd:"" group:"storage.nine.ch" name:"mysqldatabases" aliases:"mysqldatabase" help:"Get MySQL databases."`
 	Postgres            postgresCmd           `cmd:"" group:"storage.nine.ch" name:"postgres" help:"Get PostgreSQL instances."`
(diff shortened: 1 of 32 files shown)


get/mysqldatabase.go
```diff
@@ -0,0 +1,90 @@
...
+// printConnectionString according to the MySQL documentation:
+// https://dev.mysql.com/doc/refman/8.4/en/connecting-using-uri-or-key-value-pairs.html#connecting-using-uri
+func (cmd *mysqlDatabaseCmd) printConnectionString(ctx context.Context, client *api.Client, mdb *storage.MySQLDatabase) error {
+	secrets, err := getConnectionSecretMap(ctx, client, mdb)
+	if err != nil {
+		return err
+	}
+
+	for db, pw := range secrets {
+		fmt.Fprintf(cmd.out, "mysql://%s:%s@%s/%s",
```
ctrox, line 74: I would add a newline at the end of the string, else it will put the terminal prompt on the same line (depending on the shell of course). Same for the postgres equivalent.

get/get.go
```diff
@@ -24,7 +24,9 @@ type Cmd struct {
 	Releases            releasesCmd           `cmd:"" group:"deplo.io" name:"releases" aliases:"release" help:"Get deplo.io Releases."`
 	Configs             configsCmd            `cmd:"" group:"deplo.io" name:"configs" aliases:"config" help:"Get deplo.io Project Configuration."`
 	MySQL               mySQLCmd              `cmd:"" group:"storage.nine.ch" name:"mysql" help:"Get MySQL instances."`
+	MySQLDatabase       mysqlDatabaseCmd      `cmd:"" group:"storage.nine.ch" name:"mysqldatabase" help:"Get MySQL databases."`
 	Postgres            postgresCmd           `cmd:"" group:"storage.nine.ch" name:"postgres" help:"Get PostgreSQL instances."`
+	PostgresDatabase    postgresDatabaseCmd   `cmd:"" group:"storage.nine.ch" name:"postgresdatabase" help:"Get PostgreSQL databases."`
```
ctrox, line 29: Can we add some plural aliases? The first thing I tried was `get mysqldatabases`.

9marco: Thank you for the hint, it's now consistent with the other get commands (the plural being the command and singular being the alias)

PR #251, opened 2025-07-11
feat: implement copy app command

ctrox:
This implements a simple copy app command. The whole spec is copied with
some exceptions:

* Hosts are not copied to avoid conflicts with the old app unless
  explicitly forced with the `--copy-hosts` flag
* The new app is in paused state unless explicitly forced with the
  `--start` flag

The new app can be copied into a different project if desired.

### DIFF
--- a/copy/application.go
+++ b/copy/application.go
@@ -1,0 +1,106 @@
+package copy
+
+import (
+	"context"
+	"fmt"
+
+	apps "github.com/ninech/apis/apps/v1alpha1"
+	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/api/util"
+	"github.com/ninech/nctl/internal/format"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type applicationCmd struct {
+	resourceCmd
+	Start     bool `help:"Automatically start copied app."`
+	CopyHosts bool `help:"Also copy hosts of the old app."`
+}
+
+func (app *applicationCmd) Run(ctx context.Context, client *api.Client) error {
+	newApp, err := app.newCopy(ctx, client)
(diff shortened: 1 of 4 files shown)


copy/application.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	newApp := &apps.Application{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getName(app.TargetName),
+			Namespace: app.targetNamespace(client),
+		},
+		Spec: oldApp.Spec,
+	}
+	newApp.Spec.ForProvider.Paused = !app.Start
+	newApp.Spec.ForProvider.Hosts = []string{}
```
ctrox, line 55: Should we also add a flag to allow copying hosts? Or is there any inherent danger that we want to avoid?

thirdeyenick: I think we can have the flag. The DNS verification won't work by default in the new copied app, if the hosts where already verified in the old app. After the corresponding DNS changes have been made in the DNS system the old app still needs to be deleted to make the switch possible.

ctrox: Yeah I figured as much. Added the flag now as I think it will come up at some point anyway.

thirdeyenick: wouldn't copying the secret be more safe? Changing credentials on the new copy will also lead to a change of credentials on the old app, no?

copy/application.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	if err := client.Get(ctx, client.Name(oldApp.Spec.ForProvider.Git.Auth.FromSecret.Name), secret); err != nil {
+		return err
+	}
+	newSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secret.Name,
```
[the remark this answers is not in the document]
ctrox: will also add a test for the git auth copying.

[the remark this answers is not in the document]
thirdeyenick: looks good!

PR #254, opened 2025-07-18
feat: add watch flag to get commands

ctrox:
This adds a new `--watch` flag that works on most get commands. It does a normal get and then keeps nctl running while watching the specified resource for changes. This also works in combination with output options such as yaml, json etc.

As long as the resource is using the `get.list()` API, watch will automatically work. To accomplish this, the get commands need to implement a common interface `listPrinter`. For now, watch does not work in combination with the --all-projects and --all-namespaces flags.

### DIFF
--- a/get/get.go
+++ b/get/get.go
@@ -1,3 +1,4 @@
+// Package get implements the get command.
 package get
 
 import (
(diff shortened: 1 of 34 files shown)


get/get.go
```diff
@@ -33,38 +33,77 @@ type Cmd struct {
...
+	cmd.initOut()
+	return nil
+}
+
+type listPrinter interface {
+	print(context.Context, *api.Client, runtimeclient.ObjectList, *output) error
+	list() runtimeclient.ObjectList
+}
+
+func (cmd *Cmd) list(ctx context.Context, client *api.Client, lp listPrinter, opts ...api.ListOpt) error {
```
ctrox, line 70: Should this be renamed to something like `listPrint` since it takes care of both now?

get/get.go
```diff
@@ -33,38 +33,77 @@ type Cmd struct {
 	ServiceConnection   serviceConnectionCmd  `cmd:"" group:"networking.nine.ch" name:"serviceconnection" aliases:"sc" help:"Get a ServiceConnection."`
 }
 
+type output struct {
+	Format        outputFormat `help:"Configures list output. ${enum}" name:"output" short:"o" enum:"full,no-header,contexts,yaml,stats,json" default:"full"`
+	AllProjects   bool         `help:"apply the get over all projects." short:"A" xor:"watch"`
+	AllNamespaces bool         `help:"apply the get over all namespaces." hidden:"" xor:"watch"`
+	Watch         bool         `help:"Watch resource(s) for changes and print the updated resource." short:"w" xor:"watch"`
+	tabWriter     *tabwriter.Writer
```
thde, line 41: What about:


```suggestion
     *tabwriter.Writer
```

get/get.go
```diff
@@ -33,38 +33,77 @@ type Cmd struct {
...
+	jsonOut  outputFormat = "json"
 )
 
-func (cmd *Cmd) list(ctx context.Context, client *api.Client, list runtimeclient.ObjectList, opts ...api.ListOpt) error {
+func (cmd *Cmd) AfterApply() error {
+	cmd.initOut()
+	return nil
+}
+
+type listPrinter interface {
```
thde, line 65:
```suggestion
// listPrinter needs to be implemented by all resources.
type listPrinter interface {
```

thde (approved):

get/get.go
ctrox: I had this in the beginning but kong did not like having that embedded.

thde: Ah, because it is exported. 👍🏼

PR #266, opened 2025-08-04
feat: add opensearch resource

josi19:


get/opensearch.go
```diff
@@ -0,0 +1,121 @@
...
+		return fmt.Errorf("expected %T, got %T", &storage.OpenSearchList{}, list)
+	}
+	if len(openSearchList.Items) == 0 {
+		return out.printEmptyMessage(storage.OpenSearchKind, client.Project)
+	}
+
+	if cmd.Name != "" && cmd.PrintPassword {
+		return cmd.printPassword(ctx, client, &openSearchList.Items[0], out)
+	}
+	if cmd.Name != "" && cmd.PrintUser {
```
thde, line 40: I'd do:

```go
	if cmd.Name != "" && cmd.PrintUser {
		return cmd.printSecret(get.writer, ctx, client, openSearchList.Items[0], func(db, _ string) string { return db })
	}

	if cmd.Name != "" && cmd.PrintPassword {
		return cmd.printSecret(get.writer, ctx, client, openSearchList.Items[0], func(_, pw string) string { return pw })
	}
```

get/opensearch.go
```diff
@@ -0,0 +1,121 @@
...
+		fmt.Fprintln(out.writer, storage.OpenSearchUser)
+		return nil
+	}
+	if cmd.Name != "" && cmd.PrintPort {
```
thde, line 44: Is this necessary? Because we use ingress, I think this is not even true. So I'd just remove it.

get/opensearch.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// If no indices, assume healthy
+	if len(clusterHealth.Indices) == 0 {
+		return string(worstStatus)
+	}
+	// Determine the worst status of all indices
+	for _, idx := range clusterHealth.Indices {
+		switch idx.Status {
+		case storage.OpenSearchHealthStatusRed:
+			worstStatus = storage.OpenSearchHealthStatusRed
```
thde, line 102: I can't get worse here so you can return here:

```go
return idx.Status
```

get/opensearch.go
```diff
@@ -0,0 +1,121 @@
...
+			string(openSearch.Spec.ForProvider.ClusterType),
+			openSearch.Status.AtProvider.DiskSize.String(),
+			cmd.getClusterHealth(openSearch.Status.AtProvider.ClusterHealth),
+		)
+	}
+
+	return out.tabWriter.Flush()
+}
+
+func (cmd *openSearchCmd) getClusterHealth(clusterHealth storage.OpenSearchClusterHealth) string {
```
thde, line 91: I'd not convert here:

```go
func (cmd *openSearchCmd) getClusterHealth(clusterHealth storage.OpenSearchClusterHealth) storage.OpenSearchHealthStatus {
```

get/opensearch.go
```diff
@@ -0,0 +1,121 @@
...
+	if len(clusterHealth.Indices) == 0 {
+		return string(worstStatus)
+	}
+	// Determine the worst status of all indices
+	for _, idx := range clusterHealth.Indices {
+		switch idx.Status {
+		case storage.OpenSearchHealthStatusRed:
+			worstStatus = storage.OpenSearchHealthStatusRed
+		case storage.OpenSearchHealthStatusYellow:
+			if worstStatus != storage.OpenSearchHealthStatusRed {
```
thde, line 104: As you return above, there is no need for the if clause:

```go
worstStatus = idx.Status
```

shieldwed (approved):

PR #269, opened 2025-08-25
adding flags and new funcions for the sensitive fields

SaadAssaf:
We've added new flags (--sensitive-env and --sensitive-build-env) to the create and update commands. These flags allow users to mark variables as sensitive.

Masking logic were created to redact their values (*****) in the command output.

```
./nctl get configs
PROJECT    NAME    SIZE    REPLICAS    PORT    ENVIRONMENT_VARIABLES    BASIC_AUTH    DEPLOY_JOB    AGE
test             test                                                  env_test=*****           false         <none>        14m
```

Note: We didn't apply masking for YAML or JSON output. This is by design because these formats are used by automation tools etc

### DIFF
--- a/update/application_test.go
+++ b/update/application_test.go
@@ -51,7 +51,7 @@ func TestApplication(t *testing.T) {
 					Size:            initialSize,
 					Replicas:        ptr.To(int32(1)),
 					Port:            ptr.To(int32(1337)),
-					Env:             util.EnvVarsFromMap(map[string]string{"foo": "bar"}),
+					Env:             util.EnvVarsFromMap(map[string]string{"foo": "bar", "poo": "blue"}),
 					EnableBasicAuth: ptr.To(false),
 					DeployJob: &apps.DeployJob{
 						Job: apps.Job{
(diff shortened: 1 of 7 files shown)


update/application_test.go
```diff
@@ -136,15 +136,17 @@ func TestApplication(t *testing.T) {
...
+				assert.Equal(t, util.UpdateEnvVars(existingApp.Spec.ForProvider.BuildEnv, cmd.BuildEnv, cmd.SensitiveBuildEnv, nil), updated.Spec.ForProvider.BuildEnv)
+				assert.Equal(t, "very_secret", util.EnvVarByName(updated.Spec.ForProvider.Config.Env, "SECRET_KEY").Value)
 				assert.Equal(t, *cmd.DeployJob.Command, updated.Spec.ForProvider.Config.DeployJob.Command)
 				assert.Equal(t, *cmd.DeployJob.Name, updated.Spec.ForProvider.Config.DeployJob.Name)
 				assert.Equal(t, *cmd.DeployJob.Timeout, updated.Spec.ForProvider.Config.DeployJob.Timeout.Duration)
 				assert.Equal(t, *cmd.DeployJob.Retries, *updated.Spec.ForProvider.Config.DeployJob.Retries)
 				// Retry Release/Build should be not set by default:
 				assert.Nil(t, util.EnvVarByName(updated.Spec.ForProvider.Config.Env, ReleaseTrigger))
 				assert.Nil(t, util.EnvVarByName(updated.Spec.ForProvider.BuildEnv, BuildTrigger))
+				assert.True(t, *util.EnvVarByName(updated.Spec.ForProvider.Config.Env, "SECRET_KEY").Sensitive)
```
thirdeyenick, line 149: you dereference the `Sensitive` field without checking if it is `nil` before. I also didn't see where you actually set the field to sensitive.

thirdeyenick (approved):
looking good!

PR #282, opened 2025-09-09
feat: improve client_credentials login error

ctrox:
This is especially useful in CI/CD when setting the client_id/client_secret via env.


thirdeyenick (approved):
Looks good and makes sense!

PR #327, opened 2026-01-08
feat: switch organization when project is in another org

SaadAssaf:
If set-project cannot find the project in the current organization, it tries the user's other organizations from the token and switches to the one where the project exists.

### DIFF
--- a/auth/set_project.go
+++ b/auth/set_project.go
@@ -72,9 +72,12 @@ func orgFromProject(ctx context.Context, client *api.Client, project string) (st
 		return "", fmt.Errorf("could not get user info from token: %w", err)
 	}
 
+	// Avoid unnecessary API calls if some conditions are met.
+	prefix, _, found := strings.Cut(project, "-")
+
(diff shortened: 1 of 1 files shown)


auth/set_project.go
```diff
@@ -25,7 +25,32 @@ func (s *SetProjectCmd) Run(ctx context.Context, client *api.Client) error {
 	// we get the project without using the result to be sure it exists and the
 	// user has access.
 	if err := client.Get(ctx, types.NamespacedName{Name: s.Name, Namespace: org}, &management.Project{}); err != nil {
-		if !errors.IsNotFound(err) && !errors.IsForbidden(err) {
+		if errors.IsNotFound(err) || errors.IsForbidden(err) {
+			userInfo, uiErr := api.GetUserInfoFromToken(client.Token(ctx))
+			if uiErr != nil {
+				return uiErr
```
thde: Like this we loose the initial error. You can join them together:
https://pkg.go.dev/errors#Join

PR #369, opened 2026-03-13
feat(exec): add exec support for on-demand services

thde:
Adds nctl exec sub-commands for connecting directly to on-demand services: PostgreSQL instances and databases, MySQL instances and databases, and KeyValueStore. Using their native CLI clients.

```
❯ nctl exec postgresdatabase test
psql (18.3, server 17.9 (Ubuntu 17.9-1.pgdg24.04+1))
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off, ALPN: postgresql)
Type "help" for help.

05571f2_e4416af=>
```

## New nctl exec sub-commands:

  - `nctl exec postgres <name>`  connects to a Postgres instance via `psql`
  - `nctl exec postgresdatabase <name> ` connects to a PostgresDatabase via `psql`
  - `nctl exec mysql <name>`  connects to a MySQL instance via `mysql`
  - `nctl exec mysqldatabase <name>`  connects to a MySQLDatabase via `mysql`
  - `nctl exec keyvaluestore <name>`  connects to a KeyValueStore via `redis-cli`

## Access Management

For instance-level resources (Postgres, MySQL, KeyValueStore), the command detects the caller's public IP via [ip.nine.ch](https://ip.nine.ch/), checks whether it is already in the allowed CIDR list, and if not prompts the user for confirmation before adding it. 

The resource is then patched and connectivity is awaited before launching the CLI. 

## Import from stdin

The command also supports importing using stdin like so:

```bash
mysqldump -h source.example.com -u user -ppass mydb | nctl exec mysqldatabase mydb
pg_dump -h source.example.com -U user mydb | nctl exec postgresdatabase mydb
```


thde:
TODO:
- [x] Check autocomplete support.
- [x] Add flag for DB instances to change the database.

exec/mysql.go
```diff
@@ -0,0 +1,107 @@
...
+
+	args := []string{
+		"-h", fqdn,
+		"-P", mysqlPort,
+		"-u", user,
+		"-p" + pw,
+		"--ssl-mode=REQUIRED",
+	}
+	if caPath != "" {
+		args = append(args, "--ssl-ca="+caPath)
```
pawelkuc: the CA file is passed, but `--ssl-mode=REQUIRED` means the connection is encrypted only. It does not use that CA to verify the server certificate. Is this our intention? We pass `--ssl-ca`, but also explicitly set `--ssl-mode=REQUIRED`. According to the MySQL client docs, `REQUIRED` only requires
encryption and does not verify the server certificate; CA verification requires
`VERIFY_CA`, and hostname verification requires `VERIFY_IDENTITY`. MySQL also [documents](https://dev.mysql.com/doc/refman/8.4/en/connection-options.html):
> An explicit --ssl-mode option with a value other than VERIFY_CA or VERIFY_IDENTITY, together with an explicit --ssl-ca or --ssl-capath option, produces a warning that no verification of the server certificate is performed, despite a CA certificate option being specified.

exec/postgres.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const (
+	postgresPort    = "5432"
+	postgresCommand = "psql"
+)
+
+type postgresCmd struct {
+	serviceCmd
+	Database string `name:"database" short:"d" default:"postgres" completion-predictor:"postgres_databases" help:"Database name to connect to."`
```
pawelkuc, line 23: we pass credentials via arg, so other local users/process can read these from the process list. Maybe we could have (and prefer) an env config?

same for exec/mysql.go and exec/keyvaluestore.go.

exec/postgresdatabase.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"net"
+
+	storage "github.com/ninech/apis/storage/v1alpha1"
+	"github.com/ninech/nctl/api"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type postgresDatabaseCmd struct {
+	serviceCmd
```
pawelkuc, line 15: we inherit here `AllowedCidrs` (exec/cmd.go).
But we do not implement `accessManager`, so the flag cannot do anything for `postgresDatabaseCmd`?

same for exec/mysqldatabase.go

thde: Yes, because single databases always allow all IPs. There is nothing to change there as well.

exec/mysql.go
thde: Done

exec/postgres.go
thde: I updated the logic to either use env variables or a config file if necessary.

pawelkuc (approved):

PR #384, opened 2026-04-10
feat(grafana): add commands for grafana instances

DesCenario:
This PR aims to add the necessary commands to manage Grafana instances via nctl.

PR #387, opened 2026-04-21
feat(grafana): add command

thirdeyenick:
This supersedes https://github.com/ninech/nctl/pull/384 by rebasing it and adding the `AllowLocalUsers` parameter to it. With `AllowLocalUsers`, local Grafana users are allowed to login.

### DIFF
--- a/update/grafana.go
+++ b/update/grafana.go
@@ -1,0 +1,45 @@
+package update
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	observability "github.com/ninech/apis/observability/v1alpha1"
+	"github.com/ninech/nctl/api"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type grafanaCmd struct {
+	resourceCmd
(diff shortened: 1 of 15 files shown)


update/grafana.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	observability "github.com/ninech/apis/observability/v1alpha1"
+	"github.com/ninech/nctl/api"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type grafanaCmd struct {
+	resourceCmd
+	EnableAdminAccess   *bool `help:"Give admin permissions in the Grafana instance."`
+	DisableAdminAccess  *bool `help:"Revoke admin permissions in the Grafana instance."`
```
pawelkuc: hmm... if both are set the second flag wins?
I would go with a single *bool flag here (just as we do in other places in nctl)

The same for AllowLocalUsers/DisallowLocalUsers

pawelkuc (approved):

update/grafana.go
thirdeyenick: Make sense...I changed it 👍

PR #396, opened 2026-05-04
Nctl prevent success message

SaadAssaf:


update/project_config.go
```diff
@@ -76,3 +80,22 @@ func (cmd *configCmd) applyUpdates(cfg *apps.ProjectConfig) {
 		cmd.DeployJob.applyUpdates(&cfg.Spec.ForProvider.Config)
 	}
 }
+
+func (cmd *configCmd) hasUpdates() bool {
+	if cmd.Size != nil || cmd.Port != nil || cmd.Replicas != nil || cmd.BasicAuth != nil {
```
thde: I'm not sure I like this. If we add a param, we always need to add the values here now. Isn't there a different way?

update/application.go
```diff
@@ -138,6 +142,42 @@ type dockerfileBuild struct {
 	BuildContext *string `name:"dockerfile-build-context" help:"${app_dockerfile_build_context_help}" placeholder:"."`
 }
 
+func (cmd *applicationCmd) changesGiven() bool {
```
thde: So I see two patterns:
1. `applyUpdates` with returning a `bool`.
2. A separate function `changesGiven` returning a bool.

Is there any reason why we have those two patterns and why it is not done the same way everywhere?

update/update.go
```diff
@@ -67,10 +68,16 @@ func (u *updater) Update(ctx context.Context) error {
...
 	}
 
+	oldRV := u.mg.GetResourceVersion()
+
 	if err := u.client.Update(ctx, u.mg); err != nil {
 		return err
 	}
 
+	if u.mg.GetResourceVersion() == oldRV {
+		return fmt.Errorf("no changes detected; use --help to see available flags")
```
thde: We should probably use an error with context and suggestion:
https://github.com/ninech/nctl/blob/6a0064290060e5dcf5f96e45385f249be8452f82/auth/cluster.go#L47

PR #397, opened 2026-05-04
feat: avoid historical Loki live-tail queries

pawelkuc:
Follow mode previously passed the normal log start time into Loki's websocket tail endpoint. With the default 30 day window, apps with a lot of logs could force Loki to establish a live tail over a big historical range, putting pressure on queriers and query frontends.

We now fetch the requested backlog with the range query API first, then start the websocket tail from the end of that range.


thirdeyenick (commented):
Looks good. So the `QueryRange` call query is more performant than the `TailQuery` call with longer periods of data, right?

pawelkuc:
Yes, exactly. `TailQuery` seems to be optimized for the live streaming part. I also tested this locally against the debug scenario, and the fix worked without issues there.

thirdeyenick (approved):

PR #406, opened 2026-05-27
change the help text for default buildpack stack

thirdeyenick:


create/application.go
```diff
@@ -735,6 +735,6 @@ func ApplicationKongVars() (kong.Vars, error) {
 		"named Dockerfile will be searched in the application code root directory."
 	result["app_dockerfile_build_context_help"] = "Defines the build context. If left empty, the application code root directory will be used as build context."
 	result["app_buildpack_stack_help"] = "BuildpackStack sets the stack of buildpacks to use for building the application. " +
-		"If left empty, the default stack (paketo) will be used. "
+		"If left empty, the default stack (heroku) will be used. "
```
thde, line 738: We don't expose the default on API side so we can set this dynamically?

thde (approved):

create/application.go
thirdeyenick: The default gets set in an annotation in the API. We could also expose it as a const, but then we have to take care of it in 2 places (annotation and const). Not sure this is better...

PR #407, opened 2026-06-01
feat(application): set default replicas on create, allow unsetting on update

thirdeyenick:
When creating an application, replicas was left unset, meaning a change to the API's default replica count would silently affect existing applications. Replicas are now always set explicitly to the current recommended replicas amount at creation time.

On update, --unset-replicas (mutually exclusive with --replicas) sets the field back to null, allowing the value from .deploio.yaml or other configuration layers to take effect.

### DIFF
--- a/create/application_test.go
+++ b/create/application_test.go
@@ -3,11 +3,13 @@ package create
 import (
 	"bytes"
 	"context"
+	"io"
 	"os"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/alecthomas/kong"
 	runtimev1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/grafana/loki/v3/pkg/logcli/output"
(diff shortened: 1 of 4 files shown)


create/application_test.go
```diff
@@ -837,3 +839,33 @@ func setResourceCondition(ctx context.Context, apiClient *api.Client, mg resourc
...
+	vars, err := ApplicationKongVars()
+	is.NoError(err)
+
+	defaultCmd := &applicationCmd{}
+	_, err = kong.Must(defaultCmd, vars, kong.BindTo(io.Discard, (*io.Writer)(nil))).Parse([]string{
+		`test-app`,
+		`--git-url=https://github.com/ninech/doesnotexist.git`,
+	})
+	is.NoError(err)
+	is.NotNil(defaultCmd.Replicas)
```
pawelkuc: `is.NotNil(defaultCmd.Replicas)` and `is.NotNil(explicitCmd.Replicas)` call `NotNil` on an `int32`, so assertions always pass. I think these lines can be removed.

update/application.go
```diff
@@ -32,7 +32,8 @@ type applicationCmd struct {
 	Port                    *int32            `help:"Port the app is listening on."`
 	HealthProbe             *healthProbe      `embed:"" prefix:"health-probe-"`
 	DeleteHealthProbe       *bool             `help:"Delete existing custom health probe."`
-	Replicas                *int32            `help:"Amount of replicas of the running app."`
+	Replicas                *int32            `help:"Amount of replicas of the running app." xor:"replicas"`
+	UnsetReplicas           *bool             `help:"Unset the replicas, deferring to other possible configuration layer values (.deploio.yaml, etc)." xor:"replicas"`
```
pawelkuc, line 36: maybe we can have a test confirming these replica update flags are mutually exclusive and that `--unset-replicas` parses to true?

pawelkuc (approved):

create/application_test.go
thirdeyenick: True that, I changed the field type as a last step and forgot about these 👍 I removed the lines.

update/application.go
thirdeyenick: I added a test for these 👍

PR #417, opened 2026-06-16
fix: return non-zero exit code if project is not found

### LINKED ISSUE #416: nctl get $resource $resource-name should return non-zero exit code if no project is found
For example: When trying to check if a project exists one can use `nctl get projects $query`:

```
$ nctl get projects asdf
 ✗ No "Projects" found 💥
```

But since it returns `0` as return value, it's not usable in scripts.

rnestler:
Fixes #416

### DIFF
--- a/get/project.go
+++ b/get/project.go
@@ -2,10 +2,12 @@ package get
 
 import (
 	"context"
+	"fmt"
 	"sort"
 
 	management "github.com/ninech/apis/management/v1alpha1"
 	"github.com/ninech/nctl/api"
+	"github.com/ninech/nctl/internal/cli"
 	"github.com/ninech/nctl/internal/format"
 )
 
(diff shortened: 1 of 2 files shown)


get/project.go
```diff
@@ -19,6 +21,10 @@ func (proj *projectCmd) Run(ctx context.Context, client *api.Client, get *Cmd) e
 		return err
 	}
 	if len(projectList) == 0 {
+		if proj.Name != "" {
+			return cli.ErrorWithContext(fmt.Errorf("project %q was not found", proj.Name)).
+				WithExitCode(cli.ExitUsageError)
+		}
 		return get.notFound(management.ProjectKind, "")
```
rnestler, line 28: This is to avoid erroring if the user wants to list all projects, but the list is just empty. But I just realized that this doesn't make sense, since at least the organization project *always* exists.

thde: Still fine to me to have some defensive code here 😄

thde (approved):

PR #429, opened 2026-07-13
skip required flags when creating a resource with -f

SaadAssaf:


SaadAssaf:
ticket: [1695](https://gitlab.nine.ch/ninech/platform/mono-infra/-/issues/1695)

create/application.go
```diff
@@ -62,7 +62,7 @@ type applicationCmd struct {
 }
 
 type gitConfig struct {
-	URL                   string  `required:"" help:"URL to the Git repository containing the application source. Both HTTPS and SSH formats are supported."`
+	URL                   string  `help:"URL to the Git repository containing the application source. Both HTTPS and SSH formats are supported."`
```
thde, line 65: I'm wondering, when using a file, the kind shouldn't be required?

```
nctl create -f
```

and not 

```
nctl create application -f
```

SaadAssaf: Agreed, it makes no sense to specify the Kind type and then do it again in the file; therefore, I will close this PR

PR #430, opened 2026-07-14
fix: reject -f flag combined with a resource kind on create

SaadAssaf:
`-f` is now only allowed when doing `nctl create -f` as `nctl create <type> -f`doesn't make any sense. A file already contains the types of its resources.

### DIFF
--- a/create/create.go
+++ b/create/create.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"os"
 	"time"
 
 	runtimev1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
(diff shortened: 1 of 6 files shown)


create/create.go
```diff
@@ -24,8 +23,7 @@ import (
 )
 
 type Cmd struct {
-	Filename            *os.File             `short:"f" help:"Create any resource from a yaml or json file." completion-predictor:"file"`
-	FromFile            fromFile             `cmd:"" group:"create-general" default:"1" name:"-f <file>" help:"Create any resource from a yaml or json file."`
+	FromFile            fromFile             `cmd:"" group:"create-general" default:"withargs" name:"-f <file>" help:"Create any resource from a yaml or json file."`
```
thde, line 26: Nice! We probably want to do the same for `apply`, `update` and `delete`?

thde (approved):

SaadAssaf:
ticket: [1695](https://gitlab.nine.ch/ninech/platform/mono-infra/-/issues/1695)

PR #438, opened 2026-08-06
feat(cloudvm): add support for Debian 13, codenamed trixie

thde:
The main customer facing change in this MR is Debian 13 support. It also includes some fixes and improvements:

1. Fixing errors when no key or cloud config file was supplied.
2. Allowing keys from flags and files to be combined.
3. Adding basic key validation.
4. Allowing to clear rescue keys.
5. Aligning SSH key flag names over databases and VMs. Old flag names continue to work.

### DIFF
--- a/create/authorized_keys.go
+++ b/create/authorized_keys.go
@@ -4,9 +4,14 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"os"
+	"slices"
 	"strings"
 
+	storage "github.com/ninech/apis/storage/v1alpha1"
 	"golang.org/x/crypto/ssh"
+
+	"github.com/ninech/nctl/internal/format"
 )
 
 // ParseAuthorizedKeys reads SSH public keys from r. Every key is expected on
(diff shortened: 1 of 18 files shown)


create/authorized_keys.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	return keys, nil
+}
+
+// AnyFile reports whether files holds at least one file Kong decoded. A
+// repeated file flag may hold nil entries, those do not count as passed as
+// there is nothing to read from them.
+func AnyFile(files []*os.File) bool {
+	for _, file := range files {
```
thirdeyenick: Guess `slices.ContainsFunc` could also be used

thirdeyenick (approved):

create/authorized_keys.go
thde: Updated

thde:
As agreed with @9marco we will merge https://github.com/ninech/nctl/pull/437 first and then rebase this one.

thde:
> As agreed with @9marco we will merge #437 first and then rebase this one.

We'll turn this around 😄

PR #440, opened 2026-08-07
fix(update): do not error for unchanged resources

### LINKED ISSUE #439: Upsert operations fails on no-operation
In one of our deployments we recently found out that our all nctl upsert operations failed with error code 80. For example:

```
$ nctl update postgres dagster --project pupilprod ...
✗ No changes detected 💥
Suggested actions:
  - use --help to see available flags
ERROR: Command returned exit code 80 with error 'None'.
```

Is that intentional?

Seems like a bug, like usual package managers also not error when an update doesn't do anything.

thde:
Exit with 0 if no changes were made to a resource, only report usage errors if no flags were provided.

I think we overshot in https://github.com/ninech/nctl/pull/418 whereas we always report an error even if flags are provided but there are just no changes. This MR changes this behavior where as:

1. We do not error, but just report that there were no changes to a resource.
2. We only error if no flags were provided during an update call which doesn't make much sense.

Resolves #439

### DIFF
--- a/update/update.go
+++ b/update/update.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"reflect"
 
+	"github.com/alecthomas/kong"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	infra "github.com/ninech/apis/infrastructure/v1alpha1"
 	"github.com/ninech/nctl/api"
@@ -42,6 +43,10 @@ type Cmd struct {
 type ResourceCmd struct {
 	format.Writer `kong:"-"`
 	Name          string `arg:"" completion-predictor:"resource_name" help:"Name of the resource to update."`
+
+	// flagsProvided reports whether the invocation carried at least one flag
+	// that can change the resource. It is set by AfterApply.
+	flagsProvided bool
 }
 
 // BeforeApply initializes Writer from Kong's bound [io.Writer].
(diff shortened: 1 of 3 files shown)


thirdeyenick (approved):
Looks good to me!

update/update.go
```diff
@@ -42,20 +43,52 @@ type Cmd struct {
...
+	}
+
+	for _, flag := range node.Flags {
+		// Flags carrying a default are skipped. Kong marks a value as set once it has
+		// been assigned by any mechanism, including the default itself, so a defaulted
+		// flag is indistinguishable from one the user passed.
+		if flag.Set && !flag.HasDefault {
+			cmd.flagsProvided = true
+			break
+		}
```
pawelkuc, line 78: There is on issue with flags declared with a default values.
Looks like Kong cannot distinguish (through Set) whether its value came from the user or the default.
For example, Bucket has: 
```golang
ClearLifecyclePolicies bool `... default:"false"`
```

An explicit command such as:

```bash
nctl update bucket bb4  --clear-lifecycle-policies=false
```

is reported as (dlv debug session):
```bash
Starting: /.../go/bin/dlv dap --listen=127.0.0.1:52357 --log-dest=3 from /.../nctl
DAP server listening at: 127.0.0.1:52357
Type 'dlv help' for list of commands.
 ✗ No flags provided 💥

Suggested actions:
  - use --help to see available flags
Process 68812 has exited with status 80
Detaching
```

Also: we should add some test for this, i.e. an explicitly provided default valued flag.

thirdeyenick: But in the case you mentioned there is no 'action' given in the end, right? The user does not want to clear the lifecycle policies, so no updates to the bucket will be made.

thde: I see that point. But I think then it gets too complex to differentiate betwen unchanged and unset and we should never report an error?

pawelkuc: That's a fair point. `--clear-lifecycle-policies=false` does not request an effective update.
My concern was mainly that our statement here is  "only report usage errors if no flags were provided" while technically a flag was provided here. I think the implementation is fine if the intended correction is instead "no actionable update was requested".
It might be just worth adjusting the wording and adding a test for this case so the intended behavior is explicit.

thde: https://github.com/ninech/nctl/pull/441/changes WDYT

PR #441, opened 2026-08-07
fix(update): do not error on flags set to their default values

thde:


thde:
Closing in favor of https://github.com/ninech/nctl/pull/442

PR #443, opened 2026-08-10
feat(create): retry in another location

9marco:
with mysql and postgres.

This makes use of the `LocationRestricted` status cause reported by the
corresponding webhook, instead of parsing the human readable response.

### DIFF
--- a/get/all.go
+++ b/get/all.go
@@ -137,7 +137,7 @@ func printItems(items []*unstructured.Unstructured, get Cmd, header bool) error 
 	get.AllProjects = true
 
 	if header {
-		get.writeHeader("NAME", "KIND", "GROUP")
+		get.writeHeader("NAME", "KIND", "GROUP", "LOCATION")
 	}
 	for _, item := range items {
 		get.writeTabRow(
@@ -145,10 +145,21 @@ func printItems(items []*unstructured.Unstructured, get Cmd, header bool) error 
 			item.GetName(),
 			item.GroupVersionKind().Kind,
 			item.GroupVersionKind().Group,
+			location(item),
 		)
 	}
 
 	return get.tabWriter.Flush()
+}
+
+// location returns the location of item, for ones that have one.
+func location(item *unstructured.Unstructured) string {
+	loc, found, err := unstructured.NestedString(item.Object, "spec", "forProvider", "location")
+	if err != nil || !found || loc == "" {
+		return noneText
+	}
+
+	return loc
 }
 
 func filteredListTypes(s *runtime.Scheme, kinds []string) ([]schema.GroupVersionKind, error) {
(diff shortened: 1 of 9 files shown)


9marco:
Tested with a hard-coded default location on staging, where the default already is `nine-cz42`

```diff
                         ForProvider: storage.MySQLParameters{
-                                Location:     cmd.Location,
+                                Location:     "nine-cz41",
```

Fallback is being used
```
0:# ./nctl create mysql --machine-type=nine-db-xs --help | grep location
      --location=nine-cz41        Where the MySQL instance is created. Available

1:# ./nctl create mysql --machine-type=nine-db-xs
Warning: the default location does not currently accept new MySQL resources, creating in "nine-cz42" instead. The location cannot be changed later, pass --location to choose a different one.
 ✓ created MySQL "rich-catwoman" in project "marco" 🏗
 ⠋ waiting for MySQL to be ready (19m55s) ⏳
```

`--location` option still takes precedence

```
0:# ./nctl create mysql --machine-type=nine-db-xs --location=nine-cz41
nctl: error: unable to create MySQL "champion-the-professor": admission webhook "managedresource.nine.ch" denied the request: resource in location not allowed, available locations: [nine-cz42]
```

---

`get all` showing locations in an additional column, where available

```
0:# ./nctl get all --all-projects
PROJECT  NAME             KIND               GROUP               LOCATION
marco    xxxxxxxxxxxx     APIServiceAccount  iam.nine.ch         <none>
marco    bucket-xxxxxxxx  Bucket             storage.nine.ch     nine-es34
marco    marco            Bucket             storage.nine.ch     nine-es34
marco    vikunja          BucketUser         storage.nine.ch     nine-es34
marco    mysql84          MySQL              storage.nine.ch     nine-es34
marco    postgres17       Postgres           storage.nine.ch     nine-cz42
marco    postgres18       Postgres           storage.nine.ch     nine-cz42
marco    marco-test       Project            management.nine.ch  <none>
marco    marco-ssh-key    SSHKey             security.nine.ch    <none>
```

get/all.go
```diff
@@ -137,20 +137,31 @@ func printItems(items []*unstructured.Unstructured, get Cmd, header bool) error
...
 	}
 
 	return get.tabWriter.Flush()
 }
 
+// location returns the location of item, for ones that have one.
+func location(item *unstructured.Unstructured) string {
+	loc, found, err := unstructured.NestedString(item.Object, "spec", "forProvider", "location")
+	if err != nil || !found || loc == "" {
+		return noneText
```
thde, line 159: I wonder if none or global would make more sense? Maybe I like none more.

thde (approved):
Neat! ✨

get/all.go
9marco: We use 'global' with Quotas in Cockpit, so I think it might make sense to use it here as well.

I will check where else we would need to add it for consistency, without breaking things.

9marco: Although if resources do not have a `location` field, `<none>` is closer to the truth 🤔

thde: I also think we can still change this later on if a different name makes more sense.

PR #444, opened 2026-08-10
fix: return empty list for -o yaml when nothing is found

SaadAssaf:
ticket: [1670](https://gitlab.nine.ch/ninech/platform/mono-infra/-/issues/1670)

### DIFF
--- a/get/get.go
+++ b/get/get.go
@@ -163,7 +163,7 @@ func (out *output) writeTabRow(project string, row ...string) {
 }
 
 func (out *output) notFound(kind, project string) error {
-	if out.Format == jsonOut {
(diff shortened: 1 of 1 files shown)


get/get.go
```diff
@@ -163,7 +163,7 @@ func (out *output) writeTabRow(project string, row ...string) {
 }
 
 func (out *output) notFound(kind, project string) error {
-	if out.Format == jsonOut {
+	if out.Format == jsonOut || out.Format == yamlOut {
```
thde, line 166: What about a method on Format to tell if it is machine readable?

thde (approved):

PR #448, opened 2026-08-26
feat(completion): add API field completion and revamp shell predictors

thde:
Improve the predictor package to provide structured shell completions, dynamic API resource resolution, and automatic flag help/placeholder enrichment.

- Introduce internal/apifield with known API field defaults and values, plus a Kong PostBuild hook to automatically populate placeholders and append available values to flag help texts.
- Introduce internal/apiresource to resolve command names and aliases to Kubernetes API GroupVersionKinds dynamically.
- Introduce internal/completion with namespaced predictors (local:*, client:*, apifield:*) and lazy API client initialization.
- Update all CLI verbs (create, update, exec, auth, apply, delete, edit, get, logs, copy) to use namespaced completion tags.
- Remove redundant Kong variable helper functions across create and update.
- Add comprehensive unit tests for completion, field enrichment, and resource resolution.


create/mysql.go
```diff
@@ -115,19 +114,7 @@ func (cmd *mySQLCmd) newMySQL(namespace string) (*storage.MySQL, error) {
 // create command
 func MySQLKongVars() kong.Vars {
 	result := make(kong.Vars)
-	result["mysql_machine_types"] = strings.Join(stringerSlice(storage.MySQLMachineTypes), ", ")
-	result["mysql_machine_default"] = storage.MySQLMachineTypeDefault.String()
-	result["mysql_location_options"] = strings.Join(stringSlice(storage.MySQLLocationOptions), ", ")
-	result["mysql_location_default"] = string(storage.MySQLLocationDefault)
-	result["mysql_version_default"] = string(storage.MySQLVersionDefault)
-	result["mysql_versions"] = strings.Join(stringSlice(storage.MySQLVersions), ", ")
 	result["mysql_mode"] = strings.Join(storage.MySQLModeDefault, ", ")
```
thde, line 117: Check if I forgot this.

internal/apiresource/apiresource.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+	}
+
+	return node.Name
+}
+
+// FindKind resolves a resource name (singular, plural, or hyphenated) to its Nine API [schema.GroupVersionKind].
+func FindKind(scheme *runtime.Scheme, resource string) (schema.GroupVersionKind, error) {
+	if resource != "" {
+		for gvk := range scheme.AllKnownTypes() {
```
pawelkuc, line 45: `FindKind` iterates over `scheme.AllKnownTypes()`, so it can be  nondeterministic if multiple API versions of the same kind are registered. This is fine today but maybe we could either document that assumption or make somehow selection deterministic (in case multiple API versions will be introduced at some point).

internal/completion/databases.go
```diff
@@ -0,0 +1,85 @@
...
+
+	names := make([]string, 0, len(databases))
+	for dbName := range databases {
+		names = append(names, dbName)
+	}
+	return names
+}
+
+// firstPositionalArg returns the first positional argument in args, skipping flag-value pairs.
+func firstPositionalArg(args []string) string {
```
pawelkuc: `firstPositionalArg` assumes every flag consumes the following token. But we also have Boolean flags and such Boolean flag before the resource name could therefore prevent database completion?
(This existed previously 🙂)

pawelkuc (approved):

PR #449, opened 2026-09-01
fix: nctl silently override service with same name

SaadAssaf:


thirdeyenick (approved):
LGTM 👍

internal/application/service.go
thde: I think this is also a problem. This data structure doesn't even allow to define multiple services with the same name and different type via the CLI? Did you test that?

thde (changes requested):

internal/application/service.go
```diff
@@ -2,28 +2,48 @@ package application
...
 	apps "github.com/ninech/apis/apps/v1alpha1"
 	"github.com/ninech/nctl/internal/format"
 )
 
-// ServiceMap is a map of service name to typed reference, used as a CLI flag type.
-type ServiceMap map[string]TypedReference
+// NamedServiceReference is a named reference to a service target, in the
+// form "name=kind/target-name". A slice of these is used as a CLI flag
+// type for --service: unlike a map keyed by name, it preserves multiple
+// entries that share the same name but a different kind.
```
thde:
```suggestion
// form "name=kind/target-name".
```

thde (approved):

PR #457, opened 2026-09-17
refactor(api): move apply-file operation out of the apply verb

thde:
Part of https://github.com/ninech/nctl/issues/352


api/apply.go
```diff
@@ -0,0 +1,120 @@
...
+type ApplyResult string
+
+const (
+	ApplyResultCreated ApplyResult = "created"
+	ApplyResultUpdated ApplyResult = "updated"
+)
+
+// CreateFromFile creates the object described by the YAML or JSON manifest in
+// file and returns it. file is closed before returning.
+func (c *Client) CreateFromFile(ctx context.Context, file *os.File) (*unstructured.Unstructured, error) {
```
thde, line 25: Could possibly also be an interface.

SaadAssaf: yeah i agree, `io.ReadCloser` instead of `os.File`  because it only needs Read + Close and not the full os.File?

apply/file.go
```diff
@@ -2,113 +2,28 @@ package apply
...
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/yaml"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type fromFile struct {
 	format.Writer `hidden:""`
-	Filename      *os.File `short:"f" completion-predictor:"local:file"`
+	Filename      *os.File `short:"f" required:"" completion-predictor:"local:file"`
```
SaadAssaf, line 13: I see that  there are two checks for "no file given":
 Kong's required tag here and `decodeManifest` own nil check for anything calling the api package. So do we actually need the `required` check (so it can fail faster)? Or it's not necessary?

SaadAssaf (commented):
it looks good to me

